PR #8, opened 2022-06-19
Add BEiT model

seanmor5:
Adds BEiT image transformer model. The implementation shares some similarities with BERT, so I rebased off of that. There is an additional semantic segmentation model I left out because the implementation got a little complex and I wanted to push up what was working so far.


jonatanklosko:
I'm going to close this one, the implementation got way out of date and doesn't use the shared transformer. If the model is still important feel free to reopen!

PR #42, opened 2022-08-18
Add M2M100

seanmor5:
WIP: Fixing some tests


marinac-dev:
Hey guys, how is it going? Any progress on this PR?
I would love to offer my help with this feature, as I want to use it in my project.
Please let me know if there is a way somehow I could help 😄

jonatanklosko:
Hey @marinac-dev! It's been a while so we need to rewrite the PR, but that's the easy part. The blocking part is fast tokenizer for M2M100 in huggingface/tokenizers, as they currently don't support it. According to https://github.com/huggingface/transformers/pull/10236#issuecomment-791986885 it was planned, but seems they didn't get to it.

test/bumblebee/text/m2m100_test.exs
```diff
@@ -0,0 +1,77 @@
...
+        "attention_mask" => attention_mask,
+        "decoder_input_ids" => decoder_input_ids
+      }
+
+      output = Axon.predict(model, params, input)
+
+      assert Nx.shape(output.last_hidden_state) == {1, 10, 1024}
+
+      assert_all_close(
+        output.last_hidden_state[[0..-1//1, 1..3, 1..3]],
```
msluszniak, line 35:
```suggestion
        output.last_hidden_state[[.., 1..3, 1..3]],
```

test/bumblebee/text/m2m100_test.exs
```diff
@@ -0,0 +1,77 @@
...
+        "attention_mask" => attention_mask,
+        "decoder_input_ids" => decoder_input_ids
+      }
+
+      output = Axon.predict(model, params, input)
+
+      assert Nx.shape(output.logits) == {1, 10, 128_112}
+
+      assert_all_close(
+        output.logits[[0..-1//1, 1..3, 1..3]],
```
msluszniak, line 69:
```suggestion
        output.logits[[.., 1..3, 1..3]],
```

SteffenDE:
Not sure if that helps: there's also https://github.com/guillaume-be/rust-tokenizers that has an implementation of a rust tokenizer for both M2M100 and Marian.

jonatanklosko:
There's still no fast tokenizer in hf/transformers. We can always revisit if applicable, but I'm going to close this. There are most likely better models already anyway.

aymanosman:
@jonatanklosko I'm interested in adding this model in order to run https://huggingface.co/facebook/nllb-200-distilled-600M. Where should I start?

jonatanklosko:
NLLB uses the M2M100 model, but it uses a different tokenizer, which already has a fast version, so that's great. The first step is to implement M2M100, which this PR did, but it's better to start from scratch, since then we changed a lot of internals (it should be easier now, because we share most transformer code across models). This is an encoder-decoder model, so the existing T5 or Bart implementations should be a good start. The hf/transformers implementation is [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/m2m_100/modeling_m2m_100.py). For more details on porting a model see [this](https://x.com/sean_moriarity/status/1715758666001928613) and [PRs](https://github.com/elixir-nx/bumblebee/pulls?q=is%3Apr+add+model+is%3Aclosed) for other models.

Once we have the model, we just need to figure out the API for configuring the source and target languages for the tokenizer, but I can have a look once we get there :)

PR #95, opened 2022-12-08
Fix Livebook example - use Kino.Image instead of stb_image dep

dbernheisel:
Add the `:stb_image` dependency so the example works.


jonatanklosko:
Hey @dbernheisel, good catch! We actually no longer need `StbImage`, `Kino.Image.new(result.image)` should just work, can you update to that? :)

dbernheisel:
Yep that worked. Thanks for the great example!

jonatanklosko (approved):
Prefect, thanks!

PR #97, opened 2022-12-09
Fix typos

kianmeng:
Found via `codespell -S deps` and `typos --format brief`

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -607,7 +607,7 @@ defmodule Bumblebee do
   of subsequent timesteps for model forward pass.
 
   Note that the number of `timesteps` may not match `num_steps` exactly.
-  `num_steps` parameterizes sampling points, however depending on the
+  `num_steps` parametrizes sampling points, however depending on the
   method, sampling certain points may require multiple forward passes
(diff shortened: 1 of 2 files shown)


lib/bumblebee.ex
```diff
@@ -607,7 +607,7 @@ defmodule Bumblebee do
   of subsequent timesteps for model forward pass.
 
   Note that the number of `timesteps` may not match `num_steps` exactly.
-  `num_steps` parameterizes sampling points, however depending on the
```
seanmor5, line 610: I think both of these spellings are actually acceptable

lib/bumblebee.ex
```diff
@@ -607,7 +607,7 @@ defmodule Bumblebee do
...
+  `num_steps` parametrizes sampling points, however depending on the
```
jonatanklosko, line 610:
```suggestion
  `num_steps` parameterizes sampling points, however depending on the
```

jonatanklosko (approved):
Thanks :)

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #106, opened 2022-12-13
Normalize fill-mask output token

jonatanklosko:
Some tokens are not particularly usable, for example `Ġcapital`, where corresponds to a leading space. This adds a normalization, such that the result is readable.

### DIFF
--- a/lib/bumblebee/text/fill_mask.ex
+++ b/lib/bumblebee/text/fill_mask.ex
@@ -89,7 +89,13 @@ defmodule Bumblebee.Text.FillMask do
             Nx.to_flat_list(top_scores),
             Nx.to_flat_list(top_indices),
             fn score, token_id ->
-              token = Bumblebee.Tokenizer.id_to_token(tokenizer, token_id)
+              # Certain tokenizers distinguish tokens with a leading space,
+              # so we normalize the result to a consistent string
+              token =
+                tokenizer
+                |> Bumblebee.Tokenizer.decode([token_id])
+                |> String.trim()
+
               %{score: score, token: token}
             end
           )


lib/bumblebee/text/fill_mask.ex
```diff
@@ -89,7 +89,13 @@ defmodule Bumblebee.Text.FillMask do
             Nx.to_flat_list(top_scores),
             Nx.to_flat_list(top_indices),
             fn score, token_id ->
-              token = Bumblebee.Tokenizer.id_to_token(tokenizer, token_id)
+              # Certain tokenizers distinguish tokens with a leading space,
+              # so we normalize the result to a consistent string
+              token =
+                tokenizer
+                |> Bumblebee.Tokenizer.decode([token_id])
+                |> String.trim()
```
josevalim, line 97:
```suggestion
                |> String.trim_leading()
```

?

josevalim (approved):

lib/bumblebee/text/fill_mask.ex
jonatanklosko: I don't know if there's a tokenizer that would output a trailing whitespace, but I wouldn't assume that's never the case, just to be safe :D

josevalim: The counter argument is that trim_leading is much cheaper and we could postpone until something needs trailing. Your call.

jonatanklosko: Ah yeah, we can totally wait for a use case. fwiw tokens are generally a few characters long.

PR #110, opened 2022-12-14
Add CamemBERT

seanmor5:
This was requested specifically for NER tasks. It's actually just a Roberta model, so there's no new implementation, but the config lists CamemBERT so it throws off our loader.

@jonatanklosko Let me know if you think it is worth adding an extra module of integration tests to make sure all of the mappings load correctly :)

AFAICT there is not a fast tokenizer, but I think users can get around this manually by loading in Python and using `convert_slow_to_fast` or whatever the conversion is in Python


jonatanklosko:
@seanmor5 maybe a single test for camembert-base in the roberta suite? I don't think we need to test all mappings.

As for the tokenizer, there is `tokenizer.json` in [camembert-base](https://huggingface.co/camembert-base/tree/main), so we should be able to add it?

jonatanklosko (approved):

PR #119, opened 2022-12-15
stable diffusion: add warning about system requirements

ConnorRigby:
I spent a couple days trying to get my GPU to work with Bumblebee only to find out my particular GPU doesn't have enough VRAM. This warning would have saved me a lot of time.

 
Signed-off-by: Connor Rigby <connorr@hey.com>


josevalim:
Thanks. We want to improve it but I think it represents the current state of affairs.

jonatanklosko:
We have a note about CPU after the cell that runs the model, maybe let's move it there as well?

ConnorRigby:
Just moved it, didn't reword tho.

jonatanklosko:
Thanks!

PR #121, opened 2022-12-16
Add zero-shot classification

seanmor5:
I will add tests tomorrow when I get the multi-prompt case down, but right now it seems to be working fine:

Screen Shot 2022-12-15 at 6 45 05 PM

Btw, is there a reason we hid documentation for `TokenClassification` ?

### DIFF
--- a/lib/bumblebee/text/zero_shot_classification.ex
+++ b/lib/bumblebee/text/zero_shot_classification.ex
@@ -1,24 +1,26 @@
 defmodule Bumblebee.Text.ZeroShotClassification do
   @moduledoc false
 
-  alias Bumblebee.Utils
   alias Bumblebee.Shared
 
-  def zero_shot_classification(model_info, tokenizer, opts \\ []) do
+  def zero_shot_classification(model_info, tokenizer, labels, opts \\ []) do
     %{model: model, params: params, spec: spec} = model_info
     Shared.validate_architecture!(spec, :for_sequence_classification)
 
     opts =
       Keyword.validate!(opts, [
-        :aggregation,
         :compile,
-        ignored_labels: ["O"],
+        hypothesis_template: &default_hypothesis_template/1,
         defn_options: []
       ])
 
     compile = opts[:compile]
     defn_options = opts[:defn_options]
     hypothesis_template = opts[:hypothesis_template] || (&default_hypothesis_template/1)
+
+    hypotheses = Enum.map(labels, hypothesis_template)
+
+    sequences_per_batch = length(labels)
 
     batch_size = compile[:batch_size]
     sequence_length = compile[:sequence_length]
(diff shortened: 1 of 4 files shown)


jonatanklosko:
Awesome! I think ideally we should follow the same format as image/text classification, so `%{predictions: [%{label: ..., score: ...}, ...]}`. For now I wouldn't include `prompt` and `label` in the output, since the user has the input, and for multiple they can just zip.

> Btw, is there a reason we hid documentation for `TokenClassification`?

We use `defdelegate` in the `Bumblebee.Text` module and there are all the docs :)

lib/bumblebee/text/zero_shot_classification.ex
```diff
@@ -0,0 +1,107 @@
...
+    |> Nx.Serving.client_postprocessing(fn scores, _metadata, {[prompt], labels, multi?} ->
+      # TODO: Handle the case where scores comes from multiple prompts
+      scores = Nx.to_flat_list(scores[[0..-1//1, 1]])
+
+      [%{scores: scores, labels: labels, prompt: prompt}]
+      |> Shared.normalize_output(multi?)
+    end)
+  end
+
+  defp default_hypothesis_template(label), do: "This example is #{label}"
```
jonatanklosko, line 93: Not sure how relevant, but that's what they do in hf/transformers:

```suggestion
  defp default_hypothesis_template(label), do: "This example is #{label}."
```

lib/bumblebee/text/zero_shot_classification.ex
```diff
@@ -0,0 +1,107 @@
...
+
+      {[prompt | _] = prompts, labels_and_hypothesis} =
+        texts
+        |> get_inputs(hypothesis_template)
+        |> Enum.unzip()
+
+      {labels, hypothesis} = Enum.unzip(labels_and_hypothesis)
+
+      all_inputs =
+        Bumblebee.apply_tokenizer(tokenizer, Enum.zip(prompts, hypothesis),
```
jonatanklosko, line 74: One thing to note is that here a single input becomes multiple inputs in the batch (each {prompt, hypothesis} pair). We could group them by adding an additional leading axis, however the number of hypothesis may be different for every input in the batch, so we can't really do that. Probably just documenting it is the way to go?

For Stable Diffusion if someone sets `num_images_per_prompt: 2`, the number of sub-inputs is fixed and we treat it as a single member of the batch (by adding a leading axis). I wonder if we should instead treat `num_images_per_prompt: 2` as two inputs, so for batch size of 1 we would generate each image separately. This way if someone sets 4 images and batch size of 1 it would take more time, but don't blow up the memory, and they can set batch size to 4 to generate at once. This decoupling gives more control to the user.

@seanmor5 @josevalim thoughts?

lib/bumblebee/text/zero_shot_classification.ex
```diff
@@ -0,0 +1,102 @@
...
+    hypothesis_template = opts[:hypothesis_template]
+
+    hypotheses = Enum.map(labels, hypothesis_template)
+
+    sequences_per_batch = length(labels)
+
+    batch_size =
+      if batch_size = compile[:batch_size] do
+        batch_size * sequences_per_batch
+      end
```
jonatanklosko, line 28: I think we should reflect the configured `:batch_size` in serving. Since the labels are now fixed we could add another axis, but we could also make it possible to run each pair in separation, see https://github.com/elixir-nx/bumblebee/pull/121#discussion_r1050591312.

jonatanklosko (approved):
:shipit:

lib/bumblebee/text/zero_shot_classification.ex
jonatanklosko: FTR we decided to stick to the current approach, that is `:batch_size` always refers to the serving input. It may be inflated by options such as `num_images_pre_prompt`, or the number of `labels` in this case.

PR #144, opened 2022-12-31
Support parameter files using Zip64

sitch:
I added a test for the zip64 functionality, but due to requiring a >2GB binary fixture, I think including it in git LFS storage is a bit excessive.  More than happy to remove these extra code bits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ bumblebee-*.tar
 
 # Temporary files, for example, from tests.
 /tmp/
+
+# Large (>2GB) test fixture
+/bumblebee/test/fixtures/pytorch/big_model_zip64.zip.pt
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,3 @@ bumblebee-*.tar
 
 # Temporary files, for example, from tests.
 /tmp/
-
-# Large (>2GB) test fixture
-/test/fixtures/pytorch/big_model_zip64.zip.pt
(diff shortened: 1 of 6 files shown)


sitch:
There's also a bit of an annoying warning from `unzip` about invalid datetimes...

jonatanklosko:
Hey @sitch, thanks for the PR!

We definitely should fix the warning, in our case it will produce a bunch of logs (since the archive usually has a lot of files) and it will confuse users. I see it's specifically the case where modification time/date is zero, so maybe it would make sense to special case that [here](https://github.com/akash-akya/unzip/blob/dc0bc3bc14219ddce6e5ba8f7fb3ae0d9017d7a9/lib/unzip.ex#L381) and just return the earliest date. Do you want to ask/patch upstream? Alternatively we can try a different package.

As for the test, I don't think it's worth using LFS. We should be good as long as existing tests pass and we know we can read any zip64 file :)

sitch:
@jonatanklosko I opened an issue <https://github.com/akash-akya/unzip/issues/17> to deal with the superfluous warnings.

As for the test, do you want me to keep the existing code which only runs the test if a user manually runs `python test/fixtures/pytorch/generate.py` in their local repo? Or should I gut it?

jonatanklosko:
Thanks! I don't think we need the additional test in this case; as long as we use a library with zip64 support and have tests for regular zips, it should be all good :)

sitch:
@jonatanklosko this PR is ready to merge.

`unzip` upgraded to v0.8.0, and zip64 test removed.  Thanks to @akash-akya for the speedy patch!

jonatanklosko (approved):
Perfect, thank you!

PR #148, opened 2023-01-18
Introduce explicit parameter mapping

### LINKED ISSUE #71: Remap layer names
Ideally we should use any layer names we want and then have an explicit name/pattern mapping from hf/transformers names. This way we can keep the models consistent, and also share more parts of the transformer models (currently they often use different layer naming).

jonatanklosko:
Closes #71.

Advantages:

* This removes **a lot** of duplication between models. For example, now all of the models use the same attention implementation, and even the whole transformer block.
* Certain models looked very differently, even though they were for the most part the same, that's because their reference implementations applied the same concepts, but implemented them on their own (naming, structuring and nesting the layers). Now the models are much more consistent, both in terms of layer nesting and naming, which makes it much easier to see the similarities and differences.
* We no longer need to follow an artificial layers nesting, that follows from the class composition in the Python implementations. For example, when building a model with head, we don't need to prefix the base model with something like `bert.`, we just add the head layers.
* We have full control over the parameters, which makes us much less coupled to the original implementation details. For example the original GPT2 implementation uses conv1d layers (in terms of implementation those are essentially transposed dense layers) for attention. The parameter mapping has enough flexibility that we actually slice the conv1d kernels into dense layer kernels and use the same attention implementation as all other models.

Disadvantages:

* A single file doesn't contain all of the model. This already wasn't the case and I don't consider this a major issue. Also, we could still be doing that and use consistent naming/implementation, but again I don't think we should.
* Each model needs explicit mapping to hf/transformers parameters. This also implies we use different parameter names than hf/transformers.


seanmor5 (commented):
@jonatanklosko I've just skimmed a bit, but my initial thoughts are that this is fantastic! It should also make adding new models significantly easier, and in general is much more manageable for us to maintain.

I will give a more in-depth review later

seanmor5:
Also, on requiring mappings, I don't think that's a big deal, HF actually requires mappings for non-PT models anyway, and we mostly end up hiding that fact from the user anyway. It's a consequence of depending on an external community for models, but I don't think it's a bad one

josevalim (approved):

josevalim:
Fantastic work!

lib/bumblebee/huggingface/transformers/model.ex
```diff
@@ -0,0 +1,78 @@
...
+  of layers:
+
+      %{
+        "encoder.blocks.{n}.self_attention.query" => "bert.encoder.layer.{n}.attention.self.query",
+        ...
+      }
+
+  Both param names and values for corresponding layers may not match
+  exactly, so they require further transformations. For example, the
+  convolution `"kernel"` in Axon corresponds to a transposed `"weight"`
```
seanmor5, line 51: I wonder if Axon should just make the switch to use `weight` everywhere like PyTorch. It would make the conversions a bit easier, though you'd still have to pay attention to doing the transpose in some cases

lib/bumblebee/huggingface/transformers/model.ex
```diff
@@ -0,0 +1,78 @@
...
+  from PyTorch. For most common layers such conversions are handled
+  automatically.
+
+  In some cases, particularly with model-specific layers/parameters,
+  we may need more control over param mapping. In such case, instead
```
seanmor5, line 56:
```suggestion
  we may need more control over the parameter mapping. In such cases, instead
```

lib/bumblebee/layers.ex
```diff
@@ -717,4 +719,68 @@ defmodule Bumblebee.Layers do
...
+
+  ## Options
+
+    * `:name` - layer name
+
+    * `:initializer` - initializer for the embeddings. Defaults to
+      `:zeros`
+
+  """
+  def embeddings(num_embeddings, embedding_size, opts \\ []) do
```
seanmor5, line 734: I know these are often used as embeddings, but not sure I'm sold on the name `embeddings` since it is so close to `Axon.embedding`

lib/bumblebee/layers/decoder.ex
```diff
@@ -148,6 +146,15 @@ defmodule Bumblebee.Layers.Decoder do
     end
   end
 
+  defnp append_attention_cache(key, value, attention_cache, offset, _opts \\ []) do
+    %{key: cached_key, value: cached_value} = attention_cache
+    indices = [0, Nx.as_type(offset, {:s, 64}), 0, 0]
```
seanmor5, line 151: When is offset not s64?

lib/bumblebee/huggingface/transformers/model.ex
jonatanklosko: I don't think we should be changing just because of PyTorch in this case, flax does use `kernel`. Parameter naming in particular is not much hassle, especially that in case we don't need to transpose, we just always look for `weight` instead of `kernel` by default.

lib/bumblebee/layers/decoder.ex
jonatanklosko: Offset comes from the cache, which we pass as model input and get back as model output, so Axon would automatically convert it to float either way.

lib/bumblebee/layers.ex
jonatanklosko: Some models accept `position_ids`, some hardcode `position_ids` to iota and use embedding layer, and finally some just use a parameter to learn the embeddings directly (kinda merging iota and embedding). I wasn't sure about the name for the same reason, but I didn't like the alternatives I came up with (`parameterized_embeddings`, `fixed_embeddings`, `learned_embeddings`). Happy to change if you have any other preference!

lib/bumblebee/layers/decoder.ex
seanmor5: Ahh makes sense, I have opened this issue in Axon to fix this: https://github.com/elixir-nx/axon/issues/464

PR #157, opened 2023-02-02
Add question answering serving

blackeuler:
I did some poking around, and I think this is how question answering would be added. I know I need to clean up the docs and some other things like testing but feedback would be appreciated on the overall approach. Thanks so much for the excellent library.

### DIFF
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -272,9 +272,15 @@ defmodule Bumblebee.Text do
   @spec fill_mask(Bumblebee.model_info(), Bumblebee.Tokenizer.t(), keyword()) :: Nx.Serving.t()
   defdelegate fill_mask(model_info, tokenizer, opts \\ []), to: Bumblebee.Text.FillMask
 
-  @type question_answering_input :: {String.t(), String.t()}
+  @type question_answering_input :: %{question: String.t(), context: String.t()}
   @type question_answering_output :: %{
-          answer: String.t()
+          predictions: list(question_answering_result())
+        }
+  @type question_answering_result :: %{
+          text: String.t(),
+          start: number(),
+          end: number(),
+          score: number()
         }
   @doc """
 
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -273,9 +273,11 @@ defmodule Bumblebee.Text do
   defdelegate fill_mask(model_info, tokenizer, opts \\ []), to: Bumblebee.Text.FillMask
 
   @type question_answering_input :: %{question: String.t(), context: String.t()}
+
   @type question_answering_output :: %{
           predictions: list(question_answering_result())
         }
+
   @type question_answering_result :: %{
           text: String.t(),
           start: number(),
@@ -283,12 +285,45 @@ defmodule Bumblebee.Text do
           score: number()
         }
   @doc """
-
   Builds serving for the question answering task.
 
-  The serving accepts `t:question_answering_input/0` and returns `t:question_answering_ouput/0`.
+  The serving accepts `t:question_answering_input/0` and returns
+  `t:question_answering_output/0`. A list of inputs is also supported.
 
-  The question answering task predicts an answer for a question given some context.
+  The question answering task finds the most probable answer to a
+  question within the given context text.
+
+  ## Options
+
+    * `:compile` - compiles all computations for predefined input shapes
+      during serving initialization. Should be a keyword list with the
+      following keys:
+
+        * `:batch_size` - the maximum batch size of the input. Inputs
+          are optionally padded to always match this batch size. Note
+          that the batch size refers to the number of prompts to classify,
+          while the model prediction is made for every combination of
+          prompt and label
+
+        * `:sequence_length` - the maximum input sequence length. Input
+          sequences are always padded/truncated to match that length
+
+      It is advised to set this option in production and also configure
+      a defn compiler using `:defn_options` to maximally reduce inference
+      time.
+
+    * `:defn_options` - the options for JIT compilation. Defaults to `[]`
+
+  ## Examples
+
+      {:ok, roberta} = Bumblebee.load_model({:hf, "deepset/roberta-base-squad2"})
+      {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "roberta-base"})
+
+      serving = Bumblebee.Text.question_answering(roberta, tokenizer)
+
+      input = %{question: "What's my name?", context: "My name is Sarah and I live in London."}
+      Nx.Serving.run(serving, input)
+      #=> %{results: [%{end: 16, score: 0.81039959192276, start: 11, text: "Sarah"}]}
 
   """
   @spec question_answering(
@@ -309,7 +344,8 @@ defmodule Bumblebee.Text do
   Builds serving for the zero-shot classification task.
 
   The serving accepts `t:zero_shot_classification_input/0` and returns
-  `t:zero_shot_classification_output/0`.
+  `t:zero_shot_classification_output/0`. A list of inputs is also
+  supported.
 
   The zero-shot task predicts zero-shot labels for a given sequence by
   proposing each label as a premise-hypothesis pairing.
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -272,6 +272,28 @@ defmodule Bumblebee.Text do
   @spec fill_mask(Bumblebee.model_info(), Bumblebee.Tokenizer.t(), keyword()) :: Nx.Serving.t()
   defdelegate fill_mask(model_info, tokenizer, opts \\ []), to: Bumblebee.Text.FillMask
 
+  @type question_answering_input :: {String.t(), String.t()}
+  @type question_answering_output :: %{
+          answer: String.t()
+        }
+  @doc """
+
+  Builds serving for the question answering task.
+
+  The serving accepts `t:question_answering_input/0` and returns `t:question_answering_ouput/0`.
+
+  The question answering task predicts an answer for a question given some context.
+
+  """
+  @spec question_answering(
+          Bumblebee.model_info(),
+          Bumblebee.Tokenizer.t(),
+          list(String.t()),
+          keyword()
+        ) :: Nx.Serving.t()
+  defdelegate question_answering(model_info, tokenizer, opts \\ []),
+    to: Bumblebee.Text.QuestionAnswering
+
   @type zero_shot_classification_input :: String.t()
   @type zero_shot_classification_output :: %{
           predictions: list(zero_shot_classification_prediction())
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -288,7 +288,6 @@ defmodule Bumblebee.Text do
   @spec question_answering(
           Bumblebee.model_info(),
           Bumblebee.Tokenizer.t(),
-          list(String.t()),
           keyword()
         ) :: Nx.Serving.t()
   defdelegate question_answering(model_info, tokenizer, opts \\ []),
--- a/lib/bumblebee/text/question_answering.ex
+++ b/lib/bumblebee/text/question_answering.ex
@@ -1,5 +1,4 @@
 defmodule Bumblebee.Text.QuestionAnswering do
-  alias Bumblebee.Tokenizer
   alias Bumblebee.Shared
   alias Bumblebee.Utils
   alias Axon
@@ -34,8 +33,7 @@ defmodule Bumblebee.Text.QuestionAnswering do
     {_init_fun, predict_fun} = Axon.build(model)
 
     scores_fun = fn params, input ->
-      # input = Utils.Nx.composite_flatten_batch(input)
-      output = predict_fun.(params, input)
+       predict_fun.(params, input)
     end
 
     Nx.Serving.new(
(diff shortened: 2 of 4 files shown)


lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,79 @@
...
+            {:error, "expected a 1-dimensional tensor or {context,text}, got: #{inspect(other)}"}
+        end)
+
+      inputs = Bumblebee.apply_tokenizer(tokenizer, inputs)
+
+      {Nx.Batch.concatenate([inputs]), {inputs, multi?}}
+    end)
+    |> Nx.Serving.client_postprocessing(fn outputs, metadata, {inputs, multi?} ->
+      answer_start_index = outputs.start_logits |> Nx.argmax() |> Nx.to_number()
+      answer_end_index = outputs.end_logits |> Nx.argmax() |> Nx.to_number()
```
jonatanklosko, line 69: Note that outputs are batched, so if a list of inputs is given, we want to return a list of outputs. Therefore the argmax should be along `axis: -1` and then we convert the batched results to flat lists, zip and produce one result for each pair.

Eventually we should support `:top_k`, in which case we need to find the most probable pairs (as [here](https://github.com/huggingface/transformers/blob/92ce53aab859012f7714dae6d6fce7a7d701e75f/src/transformers/pipelines/question_answering.py#L95)), but we can assume `top_k: 1` for now and extend separately :)

lib/bumblebee/text.ex
```diff
@@ -272,6 +272,28 @@ defmodule Bumblebee.Text do
...
+  The question answering task predicts an answer for a question given some context.
+
+  """
+  @spec question_answering(
+          Bumblebee.model_info(),
+          Bumblebee.Tokenizer.t(),
+          list(String.t()),
+          keyword()
+        ) :: Nx.Serving.t()
+  defdelegate question_answering(model_info, tokenizer, opts \\ []),
```
jonatanklosko, line 294: We should also have a basic integration test, [an example](https://github.com/elixir-nx/bumblebee/blob/main/test/bumblebee/text/text_classification_test.exs).

jonatanklosko (commented):
Thanks for working on this! I dropped some high-level feedback :)

lib/bumblebee/text/question_answering.ex
jonatanklosko: Ah, now that you do `Utils.Nx.batch_to_list`, we no longer need `axis: -1`

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
...
+
+          %{context: context, question: text} when is_binary(text) and is_binary(context) ->
+            {:ok, {context, text}}
+
+          other ->
+            {:error, "expected a 1-dimensional tensor or {context,text}, got: #{inspect(other)}"}
+        end)
+
+      inputs = Bumblebee.apply_tokenizer(tokenizer, inputs)
+      IO.inspect(inputs)
```
jonatanklosko, line 63:
```suggestion
```

test/bumblebee/text/question_answering_test.ex
```diff
@@ -0,0 +1,31 @@
...
+      }
+
+      assert %{
+               predictions: [
+                 %{
+                   text: answers,
+                   start: answer_start_index,
+                   end: answer_end_index
+                 }
+                 | rest
```
jonatanklosko, line 26:
```suggestion
```

test/bumblebee/text/question_answering_test.ex
```diff
@@ -0,0 +1,31 @@
...
+
+      text_and_context = %{
+        question: "What is your favirote color",
+        context: "My favriote color is blue"
+      }
+
+      assert %{
+               predictions: [
+                 %{
+                   text: answers,
```
jonatanklosko, line 22: We should assert against the actual text, I guess `"blue"` in this case :) Same with indices

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
...
+
+      {Nx.Batch.concatenate([inputs]), {inputs, multi?}}
+    end)
+    |> Nx.Serving.client_postprocessing(fn outputs, metadata, {inputs, multi?} ->
+      %{
+        predictions:
```
jonatanklosko, line 69: It should be `results:`.

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
...
+            Utils.Nx.batch_to_list(inputs),
+            Utils.Nx.batch_to_list(outputs),
+            fn inputs, outputs ->
+              answer_start_index = outputs.start_logits |> Nx.argmax(axis: -1) |> Nx.to_number()
+
+              answer_end_index = outputs.end_logits |> Nx.argmax(axis: -1) |> Nx.to_number()
+
+              answer_tokens = inputs["input_ids"][answer_start_index..answer_end_index]
+
+              answers = Bumblebee.Tokenizer.decode(tokenizer, answer_tokens)
```
jonatanklosko, line 80:
```suggestion
              answer = Bumblebee.Tokenizer.decode(tokenizer, answer_tokens)
```

that's a single answer, each iteration of `Enum.zip_with` processes a new answer.

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
...
+
+              %{text: answers, start: answer_start_index, end: answer_end_index}
```
jonatanklosko, line 82: Note that `answer_start_index` and `answer_end_index` are token indices, but what we want here is positions in the original text. For this we need to `return_offsets: true` in the tokenizer in preprocessing and extract the position from that, similarly to what we do in token classification :)

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
...
+      {inputs, multi?} =
+        Shared.validate_serving_input!(input, fn
+          %Nx.Tensor{shape: {_}} = input ->
+            {:ok, input}
+
+          %{context: context, question: text} when is_binary(text) and is_binary(context) ->
+            {:ok, {context, text}}
+
+          other ->
+            {:error, "expected a 1-dimensional tensor or {context,text}, got: #{inspect(other)}"}
```
jonatanklosko, line 59:
```suggestion
          %{question: question, context: context} when is_binary(question) and is_binary(context) ->
            {:ok, {question, context}}

          other ->
            {:error, "expected input map with :question and :context keys, got: #{inspect(other)}"}
```

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,88 @@
+defmodule Bumblebee.Text.QuestionAnswering do
+  alias Bumblebee.Tokenizer
+  alias Bumblebee.Shared
+  alias Bumblebee.Utils
+  alias Axon
+
+  @moduledoc false
+
+  def answer_question(model_info, tokenizer, opts \\ []) do
+    %{model: model, params: params, spec: spec} = model_info
```
jonatanklosko, line 10:
```suggestion
    %{model: model, params: params, spec: spec} = model_info
    Shared.validate_architecture!(spec, :for_question_answering)
```

lib/bumblebee/text/question_answering.ex
blackeuler: Alright, I'm going to take a short break, but I am getting the following  ** (ArgumentError) cannot invoke compiled function because the given arguments do not match compiled arguments
it feels vague, thanks for the quick responses.

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,176 @@
...
+        end)
+
+      inputs =
+        Bumblebee.apply_tokenizer(tokenizer, inputs,
+          length: sequence_length,
+          return_special_tokens_mask: true,
+          return_offsets: true
+        )
+
+      {Nx.Batch.concatenate([inputs]), {inputs, multi?}}
```
jonatanklosko, line 81:
```suggestion
      all_inputs =
        Bumblebee.apply_tokenizer(tokenizer, inputs,
          length: sequence_length,
          return_special_tokens_mask: true,
          return_offsets: true
        )

      inputs = Map.take(all_inputs, ["input_ids", "attention_mask"])

      {Nx.Batch.concatenate([inputs]), {all_inputs, multi?}}
```

that's likely what the error is coming from, the tokenizer gives us `start_offsets` and `end_offsets`, and we want to use this in postprocessing, but we don't want to pass it into the model, since we compile it to only expect `input_ids` and `attentino_mask`.

blackeuler: Ahh Im close, but now im getting tensors as the end and start, do I need to normalize?

jonatanklosko: `Nx.to_number` should be enough I think.

blackeuler: Hmm Im getting this 
```
left:  %{
              predictions: [
                %{
                  text: "blue",
                  start: answer_start_index,
                  end: answer_end_index,
                  score: score
                } | rest
              ]
            }
     right: %{
              results: [%{end: #Nx.Tensor<
      s64[32]
      [0, 4, 7, 12, 17, 22, 23, 0, 0, 2, 3, 8, 13, 16, 21, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
    >, score: 0, start: #Nx.Tensor<
      s64[32]
      [0, 0, 5, 8, 13, 18, 23, 0, 0, 0, 3, 4, 9, 14, 17, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
    >, text: " "}]
            }
```

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,172 @@
...
+
+              answer_end_index = inputs["end_offsets"] |> Nx.argmax() |> Nx.to_number()
+
+              answer_tokens = inputs["input_ids"][answer_start_index..answer_end_index]
+
+              answers = Bumblebee.Tokenizer.decode(tokenizer, answer_tokens)
+
+              %{
+                text: answers,
+                start: inputs["start_offsets"],
```
jonatanklosko, line 100:
```suggestion
                start: Nx.to_number(inputs["start_offsets"][answer_start_index]),
```

something like this, except I would move it to a separate variable :)

blackeuler: Alirght last question I hope, where does the score come from ?

jonatanklosko: The model returns start/end logits, but we need to normalize it to 0-1 with softmax, you can compute start_scores and end_scores with `Axon.Activations.softmax`, then I think the score is just start score * end score for the argmax index.

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,116 @@
...
+              ending = inputs["end_offsets"][answer_end_index] |> Nx.to_number()
+              answer_tokens = inputs["input_ids"][answer_start_index..answer_end_index]
+
+              answer = Bumblebee.Tokenizer.decode(tokenizer, answer_tokens)
+
+              %{
+                text: answer,
+                start: start,
+                end: ending,
+                score: answer_start_index * answer_end_index
```
jonatanklosko, line 109: We rather need this:

```elixir
start_scores = Axon.Activations.softmax(outputs.start_logits)
start_idx = start_scores |> Nx.argmax() |> Nx.to_number()
...

score = Nx.multiply(start_scores[start_idx], end_scores[end_idx])
```

lib/bumblebee/text/question_answering.ex
```diff
@@ -0,0 +1,116 @@
...
+                |> Nx.to_number()
+
+              answer_end_index =
+                outputs.end_logits |> Axon.Activations.softmax() |> Nx.argmax() |> Nx.to_number()
+
+              start = inputs["start_offsets"][answer_start_index] |> Nx.to_number()
+              ending = inputs["end_offsets"][answer_end_index] |> Nx.to_number()
+              answer_tokens = inputs["input_ids"][answer_start_index..answer_end_index]
+
+              answer = Bumblebee.Tokenizer.decode(tokenizer, answer_tokens)
```
jonatanklosko, line 103: Instead of decoding the tokens, we should do `binary_slice` on the context string from start to end. That's better because we don't need to run the tokenizer, and more importantly decokenization may actually give slightly different result, for example if the fragment includes unknown token, or have a leading space, etc.

lib/bumblebee/text/question_answering.ex
blackeuler: Ooof, Ill get my head around this ML stuff one day. Thanks

lib/bumblebee/text/question_answering.ex
blackeuler: I know binary_slice is in the kernel, but I got compile errors when I tried to use it. I used String.slice instead, will that suffice?

jonatanklosko: String slice works with graphemes, while start/end are expressed in terms of utf-8 bytes, so we need `binary_slice`. Note that the third in `binary_slice(binary, start, size)`, in our case `size = end - start`.

jonatanklosko:
@blackeuler thanks :) I pushed some adjustments and docs, I think we are all good now.

There are some options we may want to support going forward, in particular `top_k` and `document_stride` (which will likely require changes to the tokenizers bindings). But I think we can add those as needed.

PR #159, opened 2023-02-13
Add T5 model

seanmor5:
Google's Flan-T5 model is a really good open-source alternative to GPT, and it's based on this T5. This is still failing, I think it's something to do with the relative attention bias, but will go back and look in a little bit.

I had to change a few things:

1) I added customizable layer norm because T5 uses an RMS norm rather than a layer norm. I will eventually upstream RMS norm into Axon

2) I added a relative_attention_bias option because some T5 blocks rely on relative attention bias

3) I added an `output_norm` option to control whether or not to apply the final `output_norm` for each block. T5 does not and instead applies one globally for a block


seanmor5:
T5 also does not scale queries, so I added that layer in. It's strange though we match PT and Flax until the application of softmax on the attention weights, and then we start to diverge, but I don't think there's anything wrong with Axon's softmax implementation

lib/bumblebee/layers/transformer.ex
```diff
@@ -101,17 +112,23 @@ defmodule Bumblebee.Layers.Transformer do
...
-          {hidden_state, attention, cross_attention, block_cache} =
+          relative_attention_bias = Enum.at(relative_attention_bias, idx)
+
+          {hidden_state, attention, cross_attention, block_cache, position_bias,
+           cross_attention_position_bias} =
             block(
               state.hidden_state,
               [
                 attention_mask: attention_mask,
+                position_bias: state.position_bias,
```
jonatanklosko, line 123: Once computed in the first block, we feed the bias to all other layers and pass `nil` as the options. Would it make sense to "compute" the bias before all blocks and pass downstream?

lib/bumblebee/layers/transformer.ex
```diff
@@ -162,6 +181,13 @@ defmodule Bumblebee.Layers.Transformer do
       should add FFN nodes to the given Axon node. The function also
       receives layer name prefix as the second argument.
 
+    * `:layer_norm` (required) - configuration of the layer norm operation
+      at the end of the transformer block. By default the network uses regular
+      layer normalization configured with the following options:
+
+      * `:layer_norm_epsilon` - the epsilon used by the layer normalization
+        layers. Defaults to `1.0e-5`
```
jonatanklosko, line 189: It can still be optional, since we have a default for the epsilon. Let's also change it to `:epsilon`, since it's nested under `:layer_norm` already :)

lib/bumblebee/layers/transformer.ex
```diff
@@ -162,6 +181,13 @@ defmodule Bumblebee.Layers.Transformer do
...
+
```
jonatanklosko, line 190:
```suggestion

      Alternatively a custom 2-arity function may be given. The function
      should add a normalization node to the given Axon node. The function
      also receives layer name prefix as the second argument.

```

lib/bumblebee/text/t5.ex
```diff
@@ -0,0 +1,541 @@
...
+      |> Axon.dropout(rate: spec.dropout_rate)
+
+    %{decoder_outputs | hidden_state: hidden_state}
+  end
+
+  defp ffn(hidden_state, spec, opts) do
+    name = opts[:name]
+
+    hidden_state
+    |> Layers.rms_norm(name: join(name, "layer_norm"), epsilon: spec.layer_norm_epsilon)
```
jonatanklosko, line 445: @seanmor5 this is the output norm actually, so we don't need `:output_norm`, we just need to remove this here. That's the funny part of transformer implementations, they group the layers differently, but at the end of the day they are mostly the same :D

Also, I think we can add support for `fft: [use_bias: false]` defaulting to true, and then we don't need the custom fft altogether!

jonatanklosko (commented):
Awesome, a couple comments!

lib/bumblebee/text/t5.ex
seanmor5: @jonatanklosko I missed a step in the implementation which uses a custom `ffn` based on the activation function passed, which I am adding now so we may still need the custom `ffn` in that case, will see how generic I can make it!

lib/bumblebee/layers/transformer.ex
```diff
@@ -187,6 +214,12 @@ defmodule Bumblebee.Layers.Transformer do
     * `:kernel_initializer` - initializer for kernel weights. Defaults
       to `:glorot_uniform`
 
+    * `:projection_size` - the projection size for key, value, and query
+      states per-head. In most cases, the projection size is equal to
+      `hidden_size // num_heads`. However, if set, this will project key,
+      value, and query by a factor of `projection_size * num_heads`.
+      Defaults to `nil`
```
jonatanklosko, line 221: ?

```suggestion
    * `:attention_projection_size` - the projection size for key, value,
      and query states per-head. Defaults to `div(hidden_size, num_attention_heads)`
```

I would also place it below `:hidde_size` as other sizes, but just cosmetics :)

lib/bumblebee/layers/transformer.ex
```diff
@@ -475,46 +579,68 @@ defmodule Bumblebee.Layers.Transformer do
...
     dropout_rate = opts[:dropout_rate]
 
     query_use_bias = opts[:query_use_bias]
     key_use_bias = opts[:key_use_bias]
     value_use_bias = opts[:value_use_bias]
+    output_use_bias = opts[:output_use_bias]
+
+    relative_attention_bias = opts[:relative_attention_bias]
+
+    inner_dim =
```
jonatanklosko, line 620: Nitpick: dim -> size

lib/bumblebee/text/t5.ex
jonatanklosko: Ah, so this is the output layer, but in this case they use the state **before** the normalization layer for the shortcut connection. I removed the `:output_norm` norm option and changed shortcut to use the parent, which I think is simpler :)

lib/bumblebee/layers/transformer.ex
jonatanklosko: So technically it wouldn't need to be shared, so supporting both may be better, I added an explicit option in addition to the config. It's not used for cross attention though, so I removed that part.

jonatanklosko:
@seanmor5 I made some changes, feel free to merge once everything looks good to you. The CI is failing, probably OOM or similar, since I verified that everything passes locally.

PR #161, opened 2023-02-17
Add multinomial sampling to sequence generation

### LINKED ISSUE #35: Support more text generation strategies
Currently `Bumblebee.Text.Generation.generate/5` supports only a basic greedy strategy for selecting the next token. We should add options for more sophisticated strategies, in particular beam search and sampling.

[Here](https://huggingface.co/blog/how-to-generate) is a good reference explaining various options. And [contrastive search](https://huggingface.co/blog/introducing-csearch), a more recent development.

seanmor5:
Adds multinomial sampling strategy and top-k/top-p options for adjusting the distribution.


lib/bumblebee/text/generation.ex
```diff
@@ -121,6 +121,17 @@ defmodule Bumblebee.Text.Generation do
...
+    * `:num_beams` - the number of beams to use in a beam search. If set
+      to 1, beam search will not be used. Defaults to `1`
+
+    * `:early_stopping` - whether to stop the beam search when at least
+      `:num_beams` sentences are finished per batch or not. Defaults to
+      `false`
+
+    * `:sample` - whether or not to use random sampling. Defaults to `false`
+
+    * `:prng_key` - random key to use when sampling. Defaults to `nil`
```
jonatanklosko, line 133: Since this function is pretty higher-level, what about `:seed` instead?

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +555,134 @@ defmodule Bumblebee.Text.Generation do
...
+        sequences: sequences,
+        length: length,
+        input_length: input_length
+      })
+
+    # TODO: logits warper
+
+    vocab = Nx.iota(Nx.shape(logits), axis: -1)
+    probabilities = Axon.Activations.softmax(logits)
+    next_token = Nx.Random.choice(key, vocab, probabilities)
```
josevalim, line 672:
```suggestion
    next_token = Nx.Random.choice(key, vocab, probabilities, [])
```

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +555,134 @@ defmodule Bumblebee.Text.Generation do
...
+          )
+
+        {sequences, length, finished?, inputs, params, prng_key}
+      end
+
+    sequences
+  end
+
+  defnp sample_condition(finished?, length, max_length) do
+    Nx.logical_not(Nx.logical_or(Nx.all(finished?), Nx.equal(length, max_length)))
```
josevalim, line 635:
```suggestion
    not(Nx.all(finished?) or length == max_length)
```

?

lib/bumblebee/text/generation.ex
seanmor5: Agreed, though one thing I was thinking was if we should pass this in the serving as well and return the updated key. They do accept `key` in `generate` in HF, but they do not return an updated key, so maybe just allowing a key to the serving is enough and it is assumed you will split it outside the serving

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +555,134 @@ defmodule Bumblebee.Text.Generation do
...
+        sequences: sequences,
+        length: length,
+        input_length: input_length
+      })
+
+    # TODO: logits warper
+
+    vocab = Nx.iota(Nx.shape(logits), axis: -1)
+    probabilities = Axon.Activations.softmax(logits)
+    next_token = Nx.Random.choice(key, vocab, probabilities, [])
```
josevalim, line 672: Actually, you need to choose the number of samples.

lib/bumblebee/text/generation.ex
jonatanklosko: Servings are so high-level that it seems to me the seed is for reproducibility if anything else (or to explore different result for the same prompt), and so a number is more transparent than a tensor key. Or do you think there's a scenario where you would actually use a split key?

seanmor5: Sorry, what I mean is currently you'd need to create a new serving with a new seed to get different inputs. Maybe this is desirable though, but I'm not sure as if you want something non-deterministic you may want the option to pass a key to the serving

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +540,147 @@ defmodule Bumblebee.Text.Generation do
...
+        ) do
+    pad_token_id = opts[:pad_token_id]
+    eos_token_id = opts[:eos_token_id]
+
+    key = Nx.Random.split(prng_key)
+    {key, key_next} = {key[1], key[0]}
+
+    model_outputs = predict_fun.(params, inputs)
+
+    logits = model_outputs.logits[[0..-1//1, -1]]
```
msluszniak, line 644:
```suggestion
    logits = model_outputs.logits[[.., -1]]
```

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +540,147 @@ defmodule Bumblebee.Text.Generation do
...
+
+    {sequences, length + 1, next_finished?, inputs, key_next}
+  end
+
+  deftransformp batched_choice(key, vocab, probabilities) do
+    {batch_size, _} = Nx.shape(vocab)
+
+    tokens =
+      for i <- 0..(batch_size - 1) do
+        candidate_tokens = vocab[[i, 0..-1//1]]
```
msluszniak, line 673:
```suggestion
        candidate_tokens = vocab[[i, ..]]
```

lib/bumblebee/text/generation.ex
```diff
@@ -507,6 +540,147 @@ defmodule Bumblebee.Text.Generation do
...
+        candidate_probs = probabilities[[i, 0..-1//1]]
```
msluszniak, line 674:
```suggestion
        candidate_probs = probabilities[[i, ..]]
```

jonatanklosko:
@seanmor5 I finished sampling, we can do beam search separately, especially that we have contrastive search too :)

josevalim (approved):

josevalim:
> we can do beam search separately

If you think about it, every search we do is a _beam_ search.

PR #168, opened 2023-03-04
Add Marian model

### LINKED ISSUE #167: Support Marian
I am curious about what work is required to support the Marian model? From what i've read it appears to be a slight variation on the BART model. Would it be possible to use the Bumblebee implementation of BART to run a Marian model? Or would it require a lot of work to support a Marian model in Bumblebee? Thanks.

seanmor5:
Resolves #167


test/bumblebee/text/marian_test.exs
```diff
@@ -0,0 +1,112 @@
...
+            46,
+            50,
+            34,
+            36838,
+            82,
+            24,
+            43,
+            6,
+            4935,
+            15
```
jonatanklosko, line 45: Formatting hint: you can concat shorter lists `[...] ++ [...]` and should be just two lines :)

jonatanklosko:
Looks good! There is indeed no fast tokenizer, so that's the blocker for us.

test/bumblebee/text/marian_test.exs
```diff
@@ -0,0 +1,112 @@
...
+        "input_ids" => input_ids,
+        "decoder_input_ids" => decoder_input_ids
+      }
+
+      outputs = Axon.predict(model, params, inputs)
+
+      assert Nx.shape(outputs.hidden_state) == {1, 26, 512}
+
+      assert_all_close(
+        outputs.hidden_state[[0..-1//1, 1..3, 1..3]],
```
msluszniak, line 59:
```suggestion
        outputs.hidden_state[[.., 1..3, 1..3]],
```

test/bumblebee/text/marian_test.exs
```diff
@@ -0,0 +1,112 @@
...
+      inputs = %{
+        "input_ids" => input_ids
+      }
+
+      outputs = Axon.predict(model, params, inputs)
+
+      assert Nx.shape(outputs.logits) == {1, 10, 59514}
+
+      assert_all_close(
+        outputs.logits[[0..-1//1, 1..3, 1..3]],
```
msluszniak, line 84:
```suggestion
        outputs.logits[[.., 1..3, 1..3]],
```

lib/bumblebee/layers.ex
```diff
@@ -3,7 +3,7 @@ defmodule Bumblebee.Layers do
 
   import Nx.Defn
 
-  @unsupported_activations [:gelu_new, :quick_gelu]
+  @unsupported_activations [:gelu_new, :quick_gelu, :swish]
 
   @pi :math.pi()
```
msluszniak, line 8: `Nx` supports $\pi$ in `Nx.Constants`

jonatanklosko:
There's still no fast tokenizer, at this point I don't think it's likely to be added. We can always revisit if applicable, but I'm going to close this. There are most likely better models already anyway.

PR #174, opened 2023-03-14
Add max, avg, first token aggregation

seanmor5:
Adds the rest of the missing token classification aggregation strategies

### DIFF
--- a/lib/bumblebee/text/token_classification.ex
+++ b/lib/bumblebee/text/token_classification.ex
@@ -179,7 +179,7 @@ defmodule Bumblebee.Text.TokenClassification do
   defp override_token_labels([entity | entities], spec, strategy) do
     {final_group, word_entities} =
       Enum.reduce(entities, {[entity], []}, fn entity, {word_group, word_entities} ->
-        if entity[:is_subword] do
+        if entity.is_subword do
           {[entity | word_group], word_entities}
         else
           {[entity], [Enum.reverse(word_group) | word_entities]}
--- a/lib/bumblebee/text/token_classification.ex
+++ b/lib/bumblebee/text/token_classification.ex
@@ -182,37 +182,33 @@ defmodule Bumblebee.Text.TokenClassification do
         if entity[:is_subword] do
           {[entity | word_group], word_entities}
         else
-          {[entity], [word_group | word_entities]}
+          {[entity], [Enum.reverse(word_group) | word_entities]}
         end
       end)
 
     [final_group | word_entities]
-    |> Enum.flat_map(&aggregate_word(&1, spec, strategy))
     |> Enum.reverse()
+    |> Enum.flat_map(fn group -> 
+      {idx, score} = aggregate_word(group, strategy)
+      label = spec.id_to_label[idx]
+      Enum.map(group, fn entity -> %{entity | label: label, score: score} end)
+    end)
   end
 
-  defp aggregate_word(group, spec, :first) do
-    [first_entity | _] = group = Enum.reverse(group)
+  defp aggregate_word([first_entity | _], :first) do
     idx = Nx.argmax(first_entity.scores) |> Nx.to_number()
     score = Nx.to_number(Nx.squeeze(first_entity.scores[[idx]]))
-    label = spec.id_to_label[idx]
-
-    group
-    |> Enum.map(fn entity -> %{entity | label: label, score: score} end)
-    |> Enum.reverse()
+    {idx, score}
   end
 
-  defp aggregate_word(group, spec, :max) do
+  defp aggregate_word(group, :max) do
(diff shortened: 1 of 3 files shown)


lib/bumblebee/text/token_classification.ex
```diff
@@ -169,6 +176,57 @@ defmodule Bumblebee.Text.TokenClassification do
     end)
   end
 
+  defp override_token_labels([entity | entities], spec, strategy) do
+    {final_group, word_entities} =
+      Enum.reduce(entities, {[entity], []}, fn entity, {word_group, word_entities} ->
+        if entity[:is_subword] do
```
jonatanklosko, line 182: We want to assert the key is there, since we are adding it ourselves :)

```suggestion
        if entity.is_subword do
```

lib/bumblebee/text/token_classification.ex
```diff
@@ -169,6 +176,57 @@ defmodule Bumblebee.Text.TokenClassification do
...
+          {[entity | word_group], word_entities}
+        else
+          {[entity], [word_group | word_entities]}
+        end
+      end)
+
+    [final_group | word_entities]
+    |> Enum.flat_map(&aggregate_word(&1, spec, strategy))
+    |> Enum.reverse()
```
jonatanklosko, line 191: We could do Enum.reverse on every group we accumulate and on the groups, then we don't need to worry about reversing anywhere else.

lib/bumblebee/text/token_classification.ex
```diff
@@ -169,6 +176,57 @@ defmodule Bumblebee.Text.TokenClassification do
...
+  end
+
+  defp aggregate_word(group, spec, :max) do
+    max_entity = Enum.max_by(group, &Nx.to_number(Nx.argmax(&1.scores)))
+    idx = Nx.argmax(max_entity.scores) |> Nx.to_number()
+    score = Nx.to_number(Nx.squeeze(max_entity.scores[[idx]]))
+    label = spec.id_to_label[idx]
+
+    group
+    |> Enum.map(fn entity -> %{entity | label: label, score: score} end)
```
jonatanklosko, line 212: We could also return `{idx, score}` and and share the `map`.

lib/bumblebee/text/token_classification.ex
```diff
@@ -169,6 +176,57 @@ defmodule Bumblebee.Text.TokenClassification do
...
 
+  defp override_token_labels([entity | entities], spec, strategy) do
+    {final_group, word_entities} =
+      Enum.reduce(entities, {[entity], []}, fn entity, {word_group, word_entities} ->
+        if entity[:is_subword] do
+          {[entity | word_group], word_entities}
+        else
+          {[entity], [word_group | word_entities]}
+        end
+      end)
```
jonatanklosko, line 187: We could also use `Enum.chunk_while`, though it's pretty verbose.

test/bumblebee/text/token_classification_test.exs
```diff
@@ -60,6 +60,144 @@ defmodule Bumblebee.Text.TokenClassificationTest do
              } = Nx.Serving.run(serving, text)
     end
 
+    test "correctly extracts entities with :first aggregation" do
+      assert {:ok, model_info} = Bumblebee.load_model({:hf, "dslim/bert-base-NER"})
```
msluszniak, line 64: Maybe this code for testing will be better

```elixir
    for aggregation <- [:first, :max, :average] do
      test "correctly extracts entities with :#{aggregation} aggregation" do
        assert {:ok, model_info} = Bumblebee.load_model({:hf, "dslim/bert-base-NER"})
        assert {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "bert-base-cased"})

        serving =
          Bumblebee.Text.TokenClassification.token_classification(model_info, tokenizer,
            aggregation: unquote(aggregation)
          )

        text = "I went with Jane Doe to Atlanta and we talked to John Smith about Microsoft"

        assert %{entities: [jane, atlanta, john, microsoft]} = Nx.Serving.run(serving, text)

        assert %{
                 label: "PER",
                 score: _jane_score,
                 phrase: "Jane Doe",
                 start: 12,
                 end: 20
               } = jane

        assert %{
                 label: "LOC",
                 score: _atlanta_score,
                 phrase: "Atlanta",
                 start: 24,
                 end: 31
               } = atlanta

        assert %{
                 label: "PER",
                 score: _john_score,
                 phrase: "John Smith",
                 start: 49,
                 end: 59
               } = john

        assert %{
                 label: "ORG",
                 score: _microsoft_score,
                 phrase: "Microsoft",
                 start: 66,
                 end: 75
               } = microsoft
      end
    end
```

jonatanklosko (approved):

PR #178, opened 2023-03-23
Improve word aggregation with BIO tagging

jonatanklosko:
Follow up to #174.

For "Janine Doe" we get:

```
Jan   | B-PER
##ine | B-PER
Do    | I-PER
##e   | I-PER
```

When overriding word token labels we need to make sure that non-starting word tokens get the `I-` prefix, so they become a single label during further aggregation.

### DIFF
--- a/lib/bumblebee/text/token_classification.ex
+++ b/lib/bumblebee/text/token_classification.ex
@@ -193,7 +193,7 @@ defmodule Bumblebee.Text.TokenClassification do
       {idx, score} = aggregate_word(group, strategy)
       label = spec.id_to_label[idx]
 
-      for {entity, idx} <- Enum.with_index(group) do
+      Enum.with_index(group, fn entity, idx ->
         label =
           if idx == 0 do
             label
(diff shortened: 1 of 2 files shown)


lib/bumblebee/text/token_classification.ex
```diff
@@ -192,7 +192,17 @@ defmodule Bumblebee.Text.TokenClassification do
...
+      for {entity, idx} <- Enum.with_index(group) do
+        label =
+          if idx == 0 do
+            label
+          else
+            force_inside_label(label)
+          end
+
+        %{entity | label: label, score: score}
+      end
```
seanmor5, line 205: ?

```suggestion
      Enum.with_index(group, fn entity, idx ->
        label =
          if idx == 0 do
            label
          else
            force_inside_label(label)
          end

        %{entity | label: label, score: score}
      end)
```

seanmor5 (approved):

josevalim (approved):

lib/bumblebee/text/token_classification.ex
jonatanklosko: Yesss!

PR #183, opened 2023-03-29
Add option to load models from cache without outgoing traffic

jonatanklosko:
When deploying an app that points to `{:hf, repo}`, the model can be downloaded as part of the build process. In that case we don't want to make the e-tag validation requests and always read from cache.

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -194,6 +194,11 @@ defmodule Bumblebee do
           operating system. You can also configure it globally by
           setting the `BUMBLEBEE_CACHE_DIR` environment variable
 
+        * `:offline` - if `true`, only cached files are accessed and
+          missing files result in an error. You can also configure it
+          globally by setting the `BUMBLEBEE_OFFLINE` environment
+          variable to `true`
+
         * `:auth_token` - the token to use as HTTP bearer authorization
           for remote files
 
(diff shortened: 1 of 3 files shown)


lib/bumblebee.ex
```diff
@@ -194,6 +194,11 @@ defmodule Bumblebee do
           operating system. You can also configure it globally by
           setting the `BUMBLEBEE_CACHE_DIR` environment variable
 
+        * `:offline` - if `true`, only cached files are accessed and
+          missing files result in an error. You can also configure it
+          globally by setting the `BUMBLEBEE_OFFLINE` environment
+          variable to `true`
```
josevalim, line 200: Instead of `offline` plus `BUMBLEBEE_OFFLINE`, wouldn't it be better to say: `offline: "path/to/cache"`? This way we are not depending on the env vars ourselves.

If we want to keep it as a boolean, we already have an env var for the cache directory, so do we need to introduce a new one? Perhaps `from_cache: true` to force from the cache?

jonatanklosko: I don't see the issue, offline and cache directory are orthogonal configurations. It may make sense to set either of them and leave the other as default, no?

jonatanklosko: Note that when a file is cached, by default we still make a HEAD request to check the e-tag and re-download if it's invalidated.

josevalim: You are right!

josevalim (approved):

PR #185, opened 2023-03-30
Expand deployment docs

jonatanklosko:
🐈


examples/phoenix/README.md
```diff
@@ -34,6 +34,38 @@ In such scenarios, you must change the calls from `Bumblebee.load_xyz({:hf, "mic
...
+
+There's a final important detail related to parameters. With the above configuration, a model will run on the GPU, however parameters will be loaded onto the CPU (due to the default backend), so they will need to be copied onto the GPU every time. To avoid that, you want to make sure that parameters are allocated on the same device that the computation runs on. With Bumblebee, you can pass a `:backend` option when loading the model:
+
+```elixir
+Bumblebee.load_model({:hf, "microsoft/resnet"},
+  backend: {EXLA.Backend, client: EXLA.Client.default_name()}
+)
+```
+
+When loading parameters manually, you can use [`Nx.with_default_backend/2`](https://hexdocs.pm/nx/Nx.html#with_default_backend/2).
```
josevalim, line 67: Couldn’t we wrap load_model inside Nx.with_default_backend? This way we only document one approach. And I think the client option There may not be needed right? It will default to that client anyway. :)

jonatanklosko: > And I think the client option There may not be needed right?

Correct, good call!

josevalim (approved):
Beautiful!

Issue #87, opened 2022-11-30
Encapsulate generation config


jonatanklosko:
See https://github.com/huggingface/transformers/issues/18655.

Currently the model configuration may include properties that are then used as defaults for generation. This makes it less clear where the defaults come from and unnecessarily makes model configurations larger.

jonatanklosko:
Now that we have more generation strategies, I think ideally we would support something like `strategy: {:contrastive_search, top_k: ..., alpha: ...}`. Having flat options and detecting strategy based on that is rather messy. The thing to figure out is how this ties to loading the config from upstream, but I wouldn't expect anything blocking.

josevalim:
@jonatanklosko that works but we can also do `strategy: :contrastive_search` and then pick the options but that may be hard to validate/document.

PR #192, opened 2023-04-04
Add contrastive search to sequence generation

jonatanklosko:
Support for contrastive search, as mentioned in #35.


lib/bumblebee/text/generation.ex
```diff
@@ -137,6 +173,15 @@ defmodule Bumblebee.Text.Generation do
...
       generated token when `:max_length` is reached
 
+    * `:no_repeat_ngram_length` - when set, n-grams of the given length
+      can occur only once in the generated sequence
+
+    * `:top_k` - the number of highest probability vocabulary tokens
+      considered as a continuation
+
+    * `:penalty_alpha` - balances the model confidence and the degeneration
+      penalty in contrastive search
```
jonatanklosko, line 183: IMO it would be cleaner to have the strategy explicitly stated, I added it to https://github.com/elixir-nx/bumblebee/issues/87#issuecomment-1496107201.

josevalim (approved):

PR #199, opened 2023-04-14
Add Llama

seanmor5:
Adds the popular Llama model. The model builds and the parameter map is correct, though the test is not passing. I'll look deeper into why the values differ tomorrow.


lib/bumblebee/layers.ex
```diff
@@ -948,4 +948,77 @@ defmodule Bumblebee.Layers do
...
     x * weight
   end
+
+  @doc """
+  Adds a rotary embedding layer to the network.
+  """
+  def rotary_embedding(query, key, value, position_ids, dim, opts \\ []) do
+    opts = Keyword.validate!(opts, [:name, max_position_embeddings: 2048])
+
+    inv_freq = Axon.param("inv_freq", fn _ -> {div(dim, 2)} end)
```
jonatanklosko, line 958: This is a generated value, so we can create this on the fly rather than having a parameter and that should give XLA more space to optimise (?)

test/fixtures/models/llama/config.json
```diff
@@ -0,0 +1,23 @@
+{
+  "_name_or_path": "huggingface/llama-7b",
```
jonatanklosko, line 2: Ideally we wouldn't include model in the repo, can you upload it to HF? :)

lib/bumblebee/text/llama.ex
```diff
@@ -0,0 +1,334 @@
...
+        :output_hidden_states,
+        :output_attentions,
+        :num_labels,
+        :id_to_label
+      ]) ++
+      Shared.token_options(
+        pad_token_id: 0,
+        bos_token_id: 1,
+        eos_token_id: 2
+      ) ++ Shared.generation_options(forced_bos_token_id: 0, forced_eos_token_id: 2)
```
jonatanklosko, line 57: This is moved to GenerationConfig now :)

```suggestion
      ])
      Shared.token_options(
        pad_token_id: 0,
        bos_token_id: 1,
        eos_token_id: 2
      ) ++ Shared.generation_options(forced_bos_token_id: 0, forced_eos_token_id: 2)
```

lib/bumblebee/layers/transformer.ex
```diff
@@ -56,7 +56,8 @@ defmodule Bumblebee.Layers.Transformer do
       :layer_norm,
       :norm_placement,
       :output_shortcut,
-      :scale_query?
+      :scale_query?,
+      :use_rotary_embedding?
```
jonatanklosko, line 60: Let's also mention new options in the corresponding docs.

jonatanklosko:
@seanmor5 perhaps we need to bump rust tokenizers in `elixir-nx/tokenizers`?

jonatanklosko:
@philss I tried bumping rust tokenizers to 0.13.3, but I got:

```
error[E0432]: unresolved import `onig`
 --> /Users/jonatanklosko/.asdf/installs/rust/nightly/registry/src/github.com-1ecc6299db9ec823/tokenizers-0.13.3/src/utils/onig.rs:3:5
  |
3 | use onig::Regex;
  |     ^^^^ help: a similar path exists: `super::onig`

error[E0433]: failed to resolve: use of undeclared crate or module `onig`
  --> /Users/jonatanklosko/.asdf/installs/rust/nightly/registry/src/github.com-1ecc6299db9ec823/tokenizers-0.13.3/src/utils/onig.rs:12:60
   |
12 |     pub fn find_iter<'r, 't>(&'r self, inside: &'t str) -> onig::FindMatches<'r, 't> {
   |                                                            ^^^^ use of undeclared crate or module `onig`

Some errors have detailed explanations: E0432, E0433.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `tokenizers` due to 2 previous errors
```

Are they are missing feature flags in tokenizers? Adding explicit dependency on onig didn't help either, but I don't know what I'm doing :)

seanmor5:
@jonatanklosko Not sure, I got the same thing

philss:
@jonatanklosko I think the problem is that they declaring `onig` as an optional dependency here: https://github.com/huggingface/tokenizers/blob/d19bc63c6770a8ef5e59c816b653b68cca329cde/tokenizers/Cargo.toml#L44

But they are "requiring" in the code as it would be always available.
A quick-fix for your upgrade is to declare in our crate (ex_tokenizers) the dependency with the `default` feature, that is adding `onig`:

```toml
tokenizers = { version = "0.13.3", default-features = false, features = ["default"]}
```

After that, run `cargo update`.

philss:
@jonatanklosko another fix is to use the `"unstable_wasm"` feature instead of `"default"`. It is what this issue suggests: https://github.com/huggingface/tokenizers/issues/1104 
I think it may be better for us, since it includes less dependencies :grimacing:

seanmor5:
This one just needs a new tokenizers release now!

lib/bumblebee.ex
```diff
@@ -125,6 +125,10 @@ defmodule Bumblebee do
     "LayoutLMForTokenClassification" =>
       {Bumblebee.Multimodal.LayoutLm, :for_token_classification},
     "LayoutLMModel" => {Bumblebee.Multimodal.LayoutLm, :base},
+    "LlamaModel" => {Bumblebee.Text.Llama, :base},
+    "LlamaForCausalLM" => {Bumblebee.Text.Llama, :for_causal_language_modeling},
+    "LLaMAForCausalLM" => {Bumblebee.Text.Llama, :for_causal_language_modeling},
```
jonatanklosko, line 130: ?

```suggestion
```

lib/bumblebee/text/llama.ex
```diff
@@ -0,0 +1,367 @@
...
+        :output_hidden_states,
+        :output_attentions,
+        :num_labels,
+        :id_to_label
+      ]) ++
+      Shared.token_options(
+        pad_token_id: 0,
+        bos_token_id: 1,
+        eos_token_id: 2
+      )
```
jonatanklosko, line 57:
```suggestion
      ])
```

lib/bumblebee/text/llama.ex
```diff
@@ -0,0 +1,367 @@
...
+      name: join(name, "blocks")
+    )
+  end
+
+  defp gated_ffn(hidden_state, intermediate_size, output_size, opts) do
+    name = opts[:name]
+    activation = opts[:activation]
+
+    hidden_state
+    |> Axon.dense(intermediate_size, name: join(name, "gate"), use_bias: false)
```
jonatanklosko, line 302: To keep the naming consistent we can have "gate", "intermediate" and "output" as in [t5](https://github.com/elixir-nx/bumblebee/blob/dc3cfee7b6c812f50d5bdf10cc15074f1c1659e0/lib/bumblebee/text/t5.ex#L492-L523)?

lib/bumblebee/layers/transformer.ex
```diff
@@ -276,6 +283,7 @@ defmodule Bumblebee.Layers.Transformer do
         :num_attention_heads,
         :hidden_size,
         :ffn,
+        position_ids: Layers.none(),
```
jonatanklosko, line 286: Let's put this on in the docs too :D

lib/bumblebee/text/llama.ex
```diff
@@ -0,0 +1,367 @@
...
+  import Bumblebee.Utils.Model, only: [join: 2]
+
+  alias Bumblebee.Layers
+
+  @impl true
+  def architectures(),
+    do: [
+      :base,
+      :for_causal_language_modeling,
+      :for_sequence_classification
```
jonatanklosko, line 138: We are missing this one.

philss:
> This one just needs a new tokenizers release now!

@seanmor5 I'm working on that :)

philss:
@seanmor5 @jonatanklosko version `0.3.2` of [tokenizers](https://hex.pm/packages/tokenizers) was just released!

lib/bumblebee.ex
seanmor5: From trying to load:

https://huggingface.co/decapoda-research/llama-7b-hf

Not sure if it's an error or not

feynmanliang:
Is there anywhere we can warn users that Llama (and its finetunes) are GPL v3, rather than the Apache 2 that most Elixir packages come licensed as.

jonatanklosko:
> A thought on usability: how do commercial users easily discover that Llama (and its finetunes) are GPL v3 which is usually not license compatible, compared the the Apache 2.0 one usually expects in Elixir OSS?

Fair point, though I'm not sure there's anything specific we should do. A commercial user should always check the huggingface repository for license and other details. The elixir codebase is one thing, but fetching a model from Hugging Face is clearly downloading external files and should be treated as such.

Technically if someone trained llama from scratch using HF transformers I believe it could be distributed under a more permissive license.

PR #211, opened 2023-05-20
Add support for `function_to_apply` in classification

coderrg:
For classification models, users should have some way of specifying the function to be applied to the logits rather than `softmax` always being used since the logits do not always correspond to mutually exclusive classes.

Take, for example, a model's logits representing whether a piece of text conveys negative, neutral, or positive sentiment in a text classification task. In this case, the `softmax` activation function is apt as it ensures that the probabilities in the final output vector sum to 1. However, if the logits denote whether the text is toxic, obscene, insulting, etc. (as seen in [Detoxify models](https://github.com/unitaryai/detoxify)), the `sigmoid` activation function would be more appropriate since the values in the final output vector do not necessarily have to sum to 1.

HF pipelines used `softmax` as the mandatory activation function over multi-class model logits until https://github.com/huggingface/transformers/pull/8328, when they added support for the `function_to_apply` parameter which can be specified by the user.

I am proposing support for a `function_to_apply` parameter here as well. The options are the same as implemented in HF: `softmax`, `sigmoid`, and `none` (`none` applies no activation function and just returns the raw logits). `softmax` is the default since it preserves backward compatibility and seems to be the most common case.

### DIFF
--- a/lib/bumblebee/text/text_classification.ex
+++ b/lib/bumblebee/text/text_classification.ex
@@ -8,11 +8,11 @@ defmodule Bumblebee.Text.TextClassification do
     Shared.validate_architecture!(spec, :for_sequence_classification)
 
     opts =
-      Keyword.validate!(opts, [:compile, top_k: 5, function_to_apply: "softmax", defn_options: []])
+      Keyword.validate!(opts, [:compile, top_k: 5, scores_function: "softmax", defn_options: []])
 
     top_k = opts[:top_k]
     compile = opts[:compile]
-    function_to_apply = opts[:function_to_apply]
+    scores_function = opts[:scores_function]
     defn_options = opts[:defn_options]
 
     batch_size = compile[:batch_size]
@@ -27,21 +27,7 @@ defmodule Bumblebee.Text.TextClassification do
 
     scores_fun = fn params, input ->
       outputs = predict_fun.(params, input)
-
-      case function_to_apply do
-        "softmax" ->
-          Axon.Activations.softmax(outputs.logits)
-
-        "sigmoid" ->
-          Axon.Activations.sigmoid(outputs.logits)
-
-        "none" ->
-          outputs.logits
-
-        _ ->
-          raise ArgumentError,
-                "Invalid :function_to_apply option. Only 'softmax', 'sigmoid', and 'none' are accepted"
-      end
+      Shared.logits_to_scores(outputs.logits, scores_function)
     end
 
     Nx.Serving.new(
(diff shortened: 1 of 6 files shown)


lib/bumblebee/text/text_classification.ex
```diff
@@ -24,7 +27,21 @@ defmodule Bumblebee.Text.TextClassification do
...
+
+        "sigmoid" ->
+          Axon.Activations.sigmoid(outputs.logits)
+
+        "none" ->
+          outputs.logits
+
+        _ ->
+          raise ArgumentError,
+                "Invalid :function_to_apply option. Only 'softmax', 'sigmoid', and 'none' are accepted"
```
jonatanklosko, line 43:
```suggestion
        other ->
          raise ArgumentError,
                "expected :scores_function to be either of :softmax, :sigmoid or :none, got: #{inspect(other}"
```

lib/bumblebee/text/text_classification.ex
```diff
@@ -24,7 +27,21 @@ defmodule Bumblebee.Text.TextClassification do
...
+      end
```
jonatanklosko, line 44: Also, we can move this whole block to the `Shared`, can be `Shared.logits_to_scores(logits, scores_function)` :)

coderrg:
Thank you for your feedback, @jonatanklosko! I have implemented your suggested changes.

lib/bumblebee/text/text_classification.ex
```diff
@@ -6,10 +6,13 @@ defmodule Bumblebee.Text.TextClassification do
   def text_classification(model_info, tokenizer, opts \\ []) do
     %{model: model, params: params, spec: spec} = model_info
     Shared.validate_architecture!(spec, :for_sequence_classification)
-    opts = Keyword.validate!(opts, [:compile, top_k: 5, defn_options: []])
+
+    opts =
+      Keyword.validate!(opts, [:compile, top_k: 5, scores_function: "softmax", defn_options: []])
```
jonatanklosko, line 11:
```suggestion
      Keyword.validate!(opts, [:compile, top_k: 5, scores_function: :softmax, defn_options: []])
```

Same for the rest :)

jonatanklosko: And let's also document the new option in [respective places](https://github.com/elixir-nx/bumblebee/blob/776e57c6b6d06c0fed47afa26d8144c7c2541149/lib/bumblebee/text.ex#L257).

coderrg: Thanks! Adjusted as per your suggestions.

jonatanklosko (approved):
Awesome, thanks :)

Issue #206, opened 2023-04-24
Get embeddings from text


sonic182:
Is there a way of obtaining embeddings from text? for example, to extract the 768 dim's from a given text using bert model

trodrigu:
https://github.com/elixir-nx/bumblebee/issues/100 may have some things that can help

trodrigu:
https://github.com/trodrigu/text_embeddings_elixir/ as an example

rakshans1:
```python
from transformers import CLIPModel, CLIPProcessor
import torch

model = CLIPModel.from_pretrained("openai/clip-vit-base-patch32")
processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch32")

text = "Hello, world!"
inputs = processor(text=text, return_tensors="pt")
model.get_text_features(**inputs)
```

Is it possible to build servings for  `get_text_feature` / `get_image_features`?

jonatanklosko:
@rakshans1 you can do both text and image the same way as the snippet above (featurizer/tokenizer + running base model). And yeah, we will have serving for both text and image.

PR #214, opened 2023-05-26
Add text embedding serving

coderrg:
Resolves https://github.com/elixir-nx/bumblebee/issues/206

### DIFF
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -305,6 +305,66 @@ defmodule Bumblebee.Text do
   defdelegate text_classification(model_info, tokenizer, opts \\ []),
     to: Bumblebee.Text.TextClassification
 
+  @type text_embedding_input :: String.t()
+  @type text_embedding_output :: %{embedding: Nx.Tensor.t()}
+
+  @doc """
+  Builds serving for text embeddings.
+
+  The serving accepts `t:text_embedding_input/0` and returns
+  `t:text_embedding_output/0`. A list of inputs is also supported.
+
+  ## Options
+
+    * `:compile` - compiles all computations for predefined input shapes
+      during serving initialization. Should be a keyword list with the
+      following keys:
+
+        * `:batch_size` - the maximum batch size of the input. Inputs
+          are optionally padded to always match this batch size
+
+        * `:sequence_length` - the maximum input sequence length. Input
+          sequences are always padded/truncated to match that length
+
+      It is advised to set this option in production and also configure
+      a defn compiler using `:defn_options` to maximally reduce inference
+      time.
+
+    * `:output_attribute` - the attribute of the embedding model output
+    to return.
+
+    * `:defn_options` - the options for JIT compilation. Defaults to `[]`
+
+  ## Examples
+
+      {:ok, model_info} = Bumblebee.load_model({:hf, "intfloat/e5-large"})
+      {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "intfloat/e5-large"})
+
+      options = [
+        output_attribute: :pooled_state
+      ]
+
+      serving = Bumblebee.Text.TextEmbedding.text_embedding(model_info, tokenizer, options)
+
+      text = "query: Cats are cute."
+      Nx.Serving.run(serving, text)
+
+      #=> %{
+      #=>   embedding: #Nx.Tensor<
+      #=>     f32[1024]
+      #=>     EXLA.Backend<host:0, 0.124908262.1234305056.185360>
+      #=>     [-0.9789889454841614, -0.9814645051956177, -0.5015208125114441, 0.9867952466011047, 0.9917466640472412, -0.5557178258895874, -0.18618212640285492, 0.797040581703186, 0.8922086954116821, 0.7599573135375977, -0.16524426639080048, -0.8740050792694092, 0.9433475732803345, 0.7217797636985779, 0.9437620639801025, 0.4694959223270416, 0.40594056248664856, -0.20143413543701172, 0.7144518494606018, -0.8689796924591064, 0.94001305103302, 0.17163503170013428, -0.9896315932273865, 0.4455447494983673, 0.41139301657676697, 0.01911175064742565, -0.11275406181812286, -0.734498143196106, -0.6410953402519226, -0.628239095211029, -0.2570168673992157, 0.475137323141098, -0.7534396052360535, -0.9492156505584717, -0.17271563410758972, 0.9081271886825562, -0.4851466119289398, -0.9440935254096985, -0.20976334810256958, -0.684502899646759, -0.11581139266490936, 0.17509342730045319, 0.05547652021050453, 0.31042391061782837, 0.955132007598877, -0.35595986247062683, 0.016105204820632935, -0.3154579997062683, 0.9630348682403564, ...]
+      #=>   >
+      #=> }
+  """
+  @spec text_embedding(
+          Bumblebee.model_info(),
+          Bumblebee.Tokenizer.t(),
+          keyword()
+        ) :: Nx.Serving.t()
+  defdelegate text_embedding(model_info, tokenizer, opts \\ []),
+    to: Bumblebee.Text.TextEmbedding
+
   @type fill_mask_input :: String.t()
   @type fill_mask_output :: %{predictions: list(fill_mask_prediction())}
   @type fill_mask_prediction :: %{score: number(), token: String.t()}
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -315,15 +315,18 @@ defmodule Bumblebee.Text do
   `t:text_embedding_output/0`. A list of inputs is also supported.
 
   ## Options
+    * `:output_attribute` - the attribute of the model output map to
+      retrieve. When the output is a single tensor (rather than a map),
+      this option is ignored. Defaults to `:pooled_state`
 
-    * `:output_attribute` - the attribute of the embedding model output
-      with the desired embedding. Set this option to `nil` in order to
-      directly retrieve the model output rather than choosing an attribute
-      of it. Defaults to `:pooled_state`
+    * `:output_pool` - pooling to apply on top of the model output, in case
+      it is not already a pooled embedding. Supported values: `nil` and
+      `:mean`. Defaults to `nil`, which applies no pooling on top of the
+      model output.
 
-    * `embedding_functions` - a list of the functions to apply to the
-      output embedding, in order. Each element of the list should be
-      one of `:l2_normalization` or `:mean_pooling`. Defaults to `[]`
+    * `:embedding_processor` - a post-processing step to apply to the
+      embedding. Supported values: `nil` and `:l2_norm`. Defaults to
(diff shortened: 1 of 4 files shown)


lib/bumblebee/text.ex
```diff
@@ -305,6 +305,66 @@ defmodule Bumblebee.Text do
...
+
+        * `:sequence_length` - the maximum input sequence length. Input
+          sequences are always padded/truncated to match that length
+
+      It is advised to set this option in production and also configure
+      a defn compiler using `:defn_options` to maximally reduce inference
+      time.
+
+    * `:output_attribute` - the attribute of the embedding model output
+    to return.
```
jonatanklosko, line 334: We try to use consistent naming across models, so I think `:pooled_state` is a sensible default :)

```suggestion
    * `:output_attribute` - the attribute of the embedding model output
      to return. Defaults to `:pooled_state`
```

Also nitpick: let's move this as the first option in the list, since compile/defn_options are in each serving and they are the most "boring" ones.

jonatanklosko:
@coderrg awesome thanks, just one nitpick from me!

@seanmor5 you've been using embeddings, so please check if this matches what you'd expect :D

coderrg:
I appreciate the suggestions @jonatanklosko and @seanmor5 — I've implemented those changes! @seanmor5 I know folks will sometimes apply L2 normalization to their embeddings, so I added support for it; let me know if there were other functions you had in mind. And @rajrajhans, thank you for bringing that up. For models like CLIP that have a projection head, it would be helpful to have the option to directly retrieve the model output.

My guess is that for most models, the pooled state (as an attribute of the output) would be used as the embedding, so perhaps it would be best to add a non-default option `:none` for `output_attribute` to account for the projection head case (e.g., CLIP).  I've gone ahead and done this for now, but I can also make `output_attribute` optional altogether if that would be preferred. Curious to hear what other folks think.

trodrigu:
This is looking great @coderrg!  I was curious about the postprocessing and [mean pooling](https://github.com/trodrigu/text_embeddings_elixir/blob/main/sentence_transformer.livemd).  Should this be a postprocessing function or is this out of scope?

coderrg:
Thanks @trodrigu — that makes sense. I've added support for `mean_pooling` as an embedding function based on your example since it seems useful for sentence transformers. Does my implementation line up with what you'd expect?

Also, since mean pooling and other functions applied to output embeddings are not mutually exclusive, I changed `embedding_function` to `embedding_functions` so that it takes a list of functions to be applied to the output embedding in order. The supported functions are now `:l2_normalization` and `:mean_pooling`.

lib/bumblebee/text/text_embedding.ex
```diff
@@ -0,0 +1,107 @@
...
+                acc_embedding
+                |> Nx.multiply(input_mask_expanded)
+                |> Nx.sum(axes: [1])
+                |> Nx.divide(Nx.sum(input_mask_expanded, axes: [1]))
+
+              other ->
+                raise ArgumentError,
+                      "expected each element of :embedding_functions to be one of :l2_normalization or :mean_pooling, got: #{inspect(other)}"
+            end
+          end)
```
jonatanklosko, line 100: We should do this as part of the `embedding_fun`, so that it is compiled together with the model itself :)

lib/bumblebee/text/text_embedding.ex
```diff
@@ -0,0 +1,107 @@
...
+      for inputs_and_embedding <- Bumblebee.Utils.Nx.batch_to_list(inputs_and_embeddings) do
+        {inputs, embedding} = inputs_and_embedding
+
+        transformed_embedding =
+          Enum.reduce(embedding_functions, embedding, fn embedding_function, acc_embedding ->
+            case embedding_function do
+              :l2_normalization ->
+                norm = Nx.LinAlg.norm(acc_embedding, ord: 2)
+
+                if norm > 0 do
```
jonatanklosko, line 81: Note that `>` works properly on tensors only inside `defn`, so let's move this whole clause to `defnp l2_norm(tensor)` :)

jonatanklosko:
Awesome, thanks for the feedback everyone!

@coderrg I left a couple more comments, but it's looking great :D

lib/bumblebee/text/text_embedding.ex
coderrg: It looks like a `defnp normalize(tensor)` function already exists in `lib/bumblebee/utils/nx.ex` so I went ahead and made it non-private and used it. Is that be fine?

coderrg:
Thanks @jonatanklosko, I've implemented your suggestions. I also added `nil` as the default option for `:output_pool` and `:embedding_processor` — please let me know if I misunderstood that suggestion, though.

lib/bumblebee/text.ex
```diff
@@ -305,6 +305,71 @@ defmodule Bumblebee.Text do
...
+  @type text_embedding_output :: %{embedding: Nx.Tensor.t()}
+
+  @doc """
+  Builds serving for text embeddings.
+
+  The serving accepts `t:text_embedding_input/0` and returns
+  `t:text_embedding_output/0`. A list of inputs is also supported.
+
+  ## Options
+    * `:output_attribute` - the attribute of the model output map to
```
jonatanklosko, line 318:
```suggestion
  ## Options

    * `:output_attribute` - the attribute of the model output map to
```

lib/bumblebee/text.ex
```diff
@@ -305,6 +305,71 @@ defmodule Bumblebee.Text do
...
+      this option is ignored. Defaults to `:pooled_state`
+
+    * `:output_pool` - pooling to apply on top of the model output, in case
+      it is not already a pooled embedding. Supported values: `nil` and
+      `:mean`. Defaults to `nil`, which applies no pooling on top of the
+      model output.
+
+    * `:embedding_processor` - a post-processing step to apply to the
+      embedding. Supported values: `nil` and `:l2_norm`. Defaults to
+      `nil`, which applies no post-processing step to the embedding.
```
jonatanklosko, line 329: We usually don't list `nil` explicitly, it's always the case for optional options :)

```suggestion
    * `:output_pool` - pooling to apply on top of the model output, in case
      it is not already a pooled embedding. Supported values: `:mean`. By
      default no pooling is applied

    * `:embedding_processor` - a post-processing step to apply to the
      embedding. Supported values: `:l2_norm`. By default the output is
      returned as is
```

lib/bumblebee/text/text_embedding.ex
```diff
@@ -0,0 +1,121 @@
...
+        end
+
+      output =
+        case embedding_processor do
+          nil ->
+            output
+
+          :l2_norm ->
+            # If the output is the zero vector, return it without normalization
+            if !Nx.any(output) do
```
jonatanklosko, line 69: We can change `normalize/1` to check if the norm is zero, so the logic stays there :)

test/bumblebee/text/text_embedding_test.exs
```diff
@@ -0,0 +1,22 @@
...
+      {:ok, model_info} = Bumblebee.load_model({:hf, "intfloat/e5-large"})
+      {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "intfloat/e5-large"})
+
+      options = [embedding_processor: :l2_norm]
+
+      serving = Bumblebee.Text.TextEmbedding.text_embedding(model_info, tokenizer, options)
+
+      text = "query: Cats are cute."
+
+      assert Nx.shape(Nx.Serving.run(serving, text).embedding) == {1024}
```
jonatanklosko, line 19: Whenever possible we test a couple values against transformer for reference:

```suggestion
      assert %{embedding: %Nx.Tensor{} = embedding} = Nx.Serving.run(serving, text)

      assert Nx.shape(embedding) == {1024}
      
      assert_all_close(
        embedding[1..3],
        Nx.tensor([-0.9815, -0.5015,  0.9868]),
        atol: 1.0e-4
      )
```

In this case, I tested against the raw model output:

```python
from transformers import AutoModel
from transformers import AutoTokenizer

model = AutoModel.from_pretrained("intfloat/e5-large")
tokenizer = AutoTokenizer.from_pretrained("intfloat/e5-large")

inputs = tokenizer("query: Cats are cute.", return_tensors="pt")
outputs = model(**inputs)
outputs.pooler_output[:, 1:4]
```

test/bumblebee/text/text_embedding_test.exs
```diff
@@ -0,0 +1,22 @@
...
+  @moduletag model_test_tags()
+
+  describe "integration" do
+    test "returns E5 embedding for a piece of text" do
+      {:ok, model_info} = Bumblebee.load_model({:hf, "intfloat/e5-large"})
+      {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "intfloat/e5-large"})
+
+      options = [embedding_processor: :l2_norm]
+
+      serving = Bumblebee.Text.TextEmbedding.text_embedding(model_info, tokenizer, options)
```
jonatanklosko, line 15:
```suggestion
      serving = Bumblebee.Text.TextEmbedding.text_embedding(model_info, tokenizer)
```

So we can assert the values from transformers. We can test for `:l2_norm` separately perhaps asserting that the embedding sum is close to 1?

jonatanklosko (commented):
Perfect, a couple final comment and we should be good to go!

lib/bumblebee/utils/nx.ex
```diff
@@ -325,14 +325,18 @@ defmodule Bumblebee.Utils.Nx do
...
       |> Nx.pow(2)
       |> Nx.sum(axes: [-1], keep_axes: true)
       |> Nx.sqrt()
 
-    tensor / norm
+    if Nx.all(norm) do
+      Nx.divide(tensor, norm)
+    else
+      tensor
+    end
```
jonatanklosko, line 339: We normalize each "row" separately, so we need to check individual norms for 0 :)

```suggestion
    norm = Nx.select(norm == 0.0, 1.0, norm)
    
    tensor / norm
```

jonatanklosko (approved):
Awesome, thanks!

lib/bumblebee/utils/nx.ex
```diff
@@ -325,13 +325,15 @@ defmodule Bumblebee.Utils.Nx do
...
+  defn normalize(tensor) do
     norm =
       tensor
       |> Nx.pow(2)
       |> Nx.sum(axes: [-1], keep_axes: true)
       |> Nx.sqrt()
 
+    norm = Nx.select(norm == 0.0, 1.0, norm)
+    
     tensor / norm
```
jonatanklosko, line 337:
```suggestion
    norm = Nx.select(norm == 0.0, 1.0, norm)

    tensor / norm
```

lib/bumblebee/utils/nx.ex
```diff
@@ -325,13 +325,15 @@ defmodule Bumblebee.Utils.Nx do
     Nx.dot(x, [-1], batch_axes, y, [-1], batch_axes)
   end
 
-  defnp normalize(tensor) do
+  defn normalize(tensor) do
```
jonatanklosko, line 328:
```suggestion
  @doc """
  Applies L2 normalization to the last dimension of the given tensor.
  """
  defn normalize(tensor) do
```

PR #221, opened 2023-06-29
Return tensors in text embedding serving

SteffenDE:
So I've been using the text embedding serving and I've been annoyed multiple times that the output is a map `%{embedding: tensor}`. I know that it's an embedding, that's what the serving is supposed to do! So in every case I always needed to do something like `Nx.Serving.batched_run(my_serving, inputs) |> Enum.map(& &1.embedding)` to get a list of tensors to do some actual work on (e.g. cosine similarity). I see no added value in wrapping the tensors in a map.

If there is something that I don't see here, please let me know.


jonatanklosko:
@SteffenDE we return a map from all other servings, I think this is good for consistency and we can easily extend it if ever needed. If we decide to return a single value then extending it will break existing code.

SteffenDE:
Do you think adding an option to prevent this would be okay? I am thinking of either:

* a simple option like `return_type: :tensor`
* a custom mapping function that defaults to `fn embeddings, _metadata -> Enum.map(embeddings, &%{embedding: &1}) end)` that one could override

I didn't test this, but https://hexdocs.pm/nx/Nx.Serving.html#client_postprocessing/2 allows to override the serving's postprocessing function. So I guess I could just do that. It does not feel right to use undocumented functions from `Bumblebee.Utils` or `Bumblebee.Shared` though.

jonatanklosko:
> Do you think adding an option to prevent this would be okay?

I wouldn't do that, options changing the output format is generally an anti-pattern in Elixir :)

Honestly, I don't think using the `Enum.map/2` on the output is a big deal. In an app, I would encapsulate the serving call behind some function anyway, so that map would be only in a single place.

jonatanklosko:
And for experimentation, it's usual that you need to somehow format outputs of one function to feed into another. For example in the case where you get a list of tensor embeddings, it's unlikely that the next function you call expects exactly a list of tensors.

SteffenDE:
Alright. Thanks for clarifying that this is indeed intentional!

PR #224, opened 2023-07-26
Update versions on CI

grzuy:


josevalim (approved):

jonatanklosko:
Hey! We should add a condition to run the last step in only one of the cases. Let's add `slow: true` to the first item and add a condition on that :)

.github/workflows/test.yaml
```diff
@@ -13,9 +13,13 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: 1.14.0
-              otp: 24.0
-            lint: true
+              elixir: "1.15.4"
+              otp: "26.0.2"
+            lint: "lint"
+            slow: "slow"
```
jonatanklosko, line 19:
```suggestion
            lint: true
            slow: true
```

grzuy: Yeah, sorry I didn't explained rationale.

I added verbosity just because GitHub PR checks summary below will only show the value, which in the case of booleans might be confusing what it means.

Anyway.
Applying suggestion, no a big deal :-)

jonatanklosko: Ah, in that case let's add `name: main (${{ matrix.pair.elixir }}, ${{ matrix.pair.otp }})` to the job :)

grzuy: Oh, didn't know you could customize that, thanks!

jonatanklosko:
Thanks :)

grzuy:
There's still a syntax error in the job name.
Will update soon.

Thanks!

jonatanklosko:
No worries, updated on main!

grzuy:
Cool, thanks!

PR #228, opened 2023-07-28
Add support for targeting multiple sequence lengths in text servings

jonatanklosko:
TODOs

* [x] mirror in other servings
* [x] override Rust tokenizers to never pad by default


lib/bumblebee/utils/tokenizers.ex
```diff
@@ -24,13 +24,26 @@ defmodule Bumblebee.Utils.Tokenizers do
...
+          encodings
+          |> Enum.map(&Encoding.n_tokens/1)
+          |> Enum.max()
+
+        case length do
+          nil ->
+            {max_length, nil}
+
+          lengths when is_list(lengths) ->
+            lengths = Enum.sort(lengths)
```
josevalim, line 43: Can we pre-sort the list and assume it has already been sorted here?

lib/bumblebee/utils/tokenizers.ex
```diff
@@ -24,13 +24,26 @@ defmodule Bumblebee.Utils.Tokenizers do
...
+            length_boundary = Enum.find(lengths, &(&1 >= max_length)) || List.last(lengths)
```
josevalim, line 44: Maybe a custom loop would be more efficient?

lib/bumblebee/utils/tokenizers.ex
jonatanklosko: We effectively dispatch the public `Bumblebee.apply_tokenizer` function to this, so we can't assume it's pre sorted. We could document this for the caller, but that's too brittle and in this case it's going to be few elements anyway.

lib/bumblebee/utils/tokenizers.ex
jonatanklosko: Yess!

josevalim (approved):

PR #229, opened 2023-07-30
Add image embedding serving

rajrajhans:
- Adds a serving for extracting embeddings from an image using Vision models like clip, blip, vit, deit.
- Supports L2 Normalization of the output embedding via `embedding_processor` attribute, similar to `TextEmbedding`.

### DIFF
--- a/lib/bumblebee/vision.ex
+++ b/lib/bumblebee/vision.ex
@@ -134,4 +134,64 @@ defmodule Bumblebee.Vision do
         ) :: Nx.Serving.t()
   defdelegate image_to_text(model_info, featurizer, tokenizer, generation_config, opts \\ []),
     to: Bumblebee.Vision.ImageToText
+
+  @type image_embedding_input :: image()
+  @type image_embedding_output :: %{embedding: Nx.Tensor.t()}
+  @doc """
+  Builds serving for image embeddings.
+
+  The serving accepts `t:image_embedding_input/0` and returns
+  `t:image_embedding_output/0`. A list of inputs is also supported.
+
+  ## Options
+
+    * `:output_attribute` - the attribute of the model output map to
+      retrieve. When the output is a single tensor (rather than a map),
+      this option is ignored. Defaults to `:pooled_state`
+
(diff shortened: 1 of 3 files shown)


lib/bumblebee/vision.ex
```diff
@@ -134,4 +134,64 @@ defmodule Bumblebee.Vision do
...
+    * `:compile` - compiles all computations for predefined input shapes
+      during serving initialization. Should be a keyword list with the
+      following keys:
+
+        * `:batch_size` - the maximum batch size of the input. Inputs
+          are optionally padded to always match this batch size
+
+        * `:image_size` - the size of the input spatial dimensions
+
+        * `:channels` - the number of channels in the input
```
jonatanklosko, line 165: We don't need to pass these shapes, we can determine them from the model, as we do for other image servings:

https://github.com/elixir-nx/bumblebee/blob/9f62d9956c5e67a1f9cc835a53a0eccce23921e8/lib/bumblebee/vision/image_classification.ex#L41-L43

jonatanklosko (commented):
Awesome, just one minor comment and we can ship it!

lib/bumblebee/vision.ex
rajrajhans: Noted! Thanks, @jonatanklosko!

test/bumblebee/vision/image_embedding_test.exs
```diff
@@ -0,0 +1,79 @@
...
+
+      assert %{embedding: %Nx.Tensor{} = embedding} = Nx.Serving.run(serving, image)
+      assert Nx.shape(embedding) == {1000}
+
+      assert_all_close(
+        embedding[1..3],
+        Nx.tensor([0.8104, -0.0693, 0.4399]),
+        atol: 1.0e-4
+      )
+    end
```
jonatanklosko, line 77: I think we can skip this one, we usually don't duplicate tests with just compile, since those tests are time-costly :)

```suggestion
```

jonatanklosko (approved):

PR #230, opened 2023-07-31
Speed up tests

jonatanklosko:
Mostly removing EXLA-compilation from places where EXLA backend is faster for one-off runs.


jonatanklosko:
This combination of tests fails to run on CI. I thought it may be because with backend it's more involved (no fusion), but I checked and it fails without the changes or even when using EXLA globally. From the logs it's not even obvious what limit is the issue, so I will merge as this is at least faster locally.

PR #231, opened 2023-08-02
feat: supports loading .safetensors params file

### LINKED ISSUE #96: Support .safetensors deserialization/serialization
We may want to move this all the way up to Nx/Axon.

grzuy:
closes #96 

Opening proof of concept as draft while I continue working on some improvements and test coverage, and potentially any other feedback folks have :-)


lib/bumblebee.ex
```diff
@@ -525,6 +528,14 @@ defmodule Bumblebee do
     end
   end
 
+  defp loader_module(".safetensors") do
+    Bumblebee.Conversion.Safetensors
+  end
+
+  defp loader_module(_) do
+    Bumblebee.Conversion.PyTorch.Loader
```
grzuy, line 536: Might make sense for this to be renamed to `PickleLoader` maybe

lib/bumblebee.ex
```diff
@@ -525,6 +528,14 @@ defmodule Bumblebee do
     end
   end
 
+  defp loader_module(".safetensors") do
+    Bumblebee.Conversion.Safetensors
```
jonatanklosko, line 532: We don't need a separate module for safetensors, because there's no additional logic. I would make the option `:loader_fun` and for safetensors point to a private function in this module :)

lib/bumblebee.ex
jonatanklosko: It's actually a PyTorch specific format, the file is a zip and only one of the contents is pickle :)

mix.exs
```diff
@@ -42,6 +42,7 @@ defmodule Bumblebee.MixProject do
       {:torchx, github: "elixir-nx/nx", sparse: "torchx", override: true, only: [:dev, :test]},
       {:nx_image, "~> 0.1.0"},
       {:unpickler, "~> 0.1.0"},
+      {:safetensors, "~> 0.1.0"},
```
jonatanklosko, line 45: I believe this fits well into the `elixir-nx` org, would you be ok with transferring the repo there? :)

jonatanklosko:
Thanks for the PR, a couple minor comments :)

lib/bumblebee.ex
grzuy: Gotcha! :-) :+1:

lib/bumblebee.ex
grzuy: Sounds good!
Done.

mix.exs
grzuy: Of course!
Will do...

grzuy: Indeed.
Got an error he :)

Added you as admin to the repo.

grzuy: Received, thanks!

How you wanna go about Hexpm package permissions?

lib/bumblebee/conversion/pytorch.ex
```diff
@@ -24,17 +24,28 @@ defmodule Bumblebee.Conversion.PyTorch do
       between the Axon model and the PyTorch state. For more details see
       `Bumblebee.HuggingFace.Transformers.Model.params_mapping/1`
 
+    * `:loader_fun` - a 1-arity function that takes a path argument
+      and loads the params file.
+      Defaults to `Bumblebee.Conversion.PyTorch.Loader.load!/1`.
```
jonatanklosko, line 29:
```suggestion
      and loads the params file. Defaults to
      `Bumblebee.Conversion.PyTorch.Loader.load!/1`
```

jonatanklosko (approved):
LGTM!

grzuy:
FWIW I plan to explore in a follow up PR also supporting safetensors sharded params files.

mix.exs
jonatanklosko, line 45:
```suggestion
      {:safetensors, "~> 0.1.1"},
```

grzuy:
> FWIW I plan to explore in a follow up PR also supporting safetensors sharded params files.

Mmm, correcting myself.

I think with the changes in this PR one would be able to load sharded safetensors param files.

For example, for 

https://huggingface.co/stabilityai/StableBeluga-7B/tree/main

which contains the following files
```
model-00001-of-00002.safetensors
model-00002-of-00002.safetensors
model.safetensors.index.json
```

using

```
Bumblebee.load_model({:hf, "stabilityai/StableBeluga-7B"}, params_filename: "model.safetensors")
```

should use the existing sharded params loading logic, look at the index file and "just work".

grzuy:
What would be a good improvement might be addind some decent auto-download of the preferred format based on what's available in the model repo without having the user needed to explicitly provide the file name.

jonatanklosko:
Good call, so far most repos had the pytorch file and optionally other formats, but as safetensors become more popular there may be cases where it's just safetensors. Currently we do fallbacks, that is, request one file, if doesn't exist request another, and so on. I checked and looks like HF API now allows listing files, so I will later reevaluate if we can improve :)

jonatanklosko:
FTR as of #256 we automatically detect if there are no parameters in the pytorch format, but safetensors one is available :)

PR #236, opened 2023-08-30
Support chunking to enable long-form transcription

jonatanklosko:
Awaits changes to Nx to support batched_run with chunks > single batch size.


lib/bumblebee/audio.ex
```diff
@@ -26,6 +26,19 @@ defmodule Bumblebee.Audio do
 
   ## Options
 
+    * `:chunk_num_seconds` - enables long-form transcription by splitting
```
josevalim, line 29: I wonder if we could have a more descriptive name, unless this is also what the Py libraries use. But I don't have any ideas. :D

josevalim: Maybe "chunk_after_seconds: ...", your call.

josevalim (approved):

lib/bumblebee/audio.ex
josevalim: Sounds good, your call!

PR #242, opened 2023-09-12
Add support for streaming speech-to-text results

jonatanklosko:
Large audio is split into multiple chunks, now we support `stream: true`, so that output chunks are emitted as soon as they are available.

### DIFF
--- a/lib/bumblebee/audio.ex
+++ b/lib/bumblebee/audio.ex
@@ -6,7 +6,27 @@ defmodule Bumblebee.Audio do
   # TODO: remove in v0.5
   @deprecated "Use Bumblebee.Audio.speech_to_text_whisper/5 instead."
   def speech_to_text(model_info, featurizer, tokenizer, generation_config, opts \\ []) do
-    speech_to_text_whisper(model_info, featurizer, tokenizer, generation_config, opts)
+    serving = speech_to_text_whisper(model_info, featurizer, tokenizer, generation_config, opts)
+    client_postprocessing = serving.client_postprocessing
+
+    Nx.Serving.client_postprocessing(serving, fn output_pair, info ->
+      output = client_postprocessing.(output_pair, info)
+
+      if is_list(output) do
+        Enum.map(output, &speech_to_text_convert_output/1)
+      else
+        speech_to_text_convert_output(output)
+      end
+    end)
+  end
+
+  defp speech_to_text_convert_output(%{chunks: chunks}) do
+    text =
+      chunks
+      |> Enum.map_join(& &1.text)
+      |> String.trim()
+
+    %{results: [%{text: text}]}
   end
 
   @typedoc """
@@ -21,15 +41,13 @@ defmodule Bumblebee.Audio do
 
   """
   @type speech_to_text_whisper_input :: Nx.t() | {:file, String.t()}
-  @type speech_to_text_whisper_output :: %{results: list(speech_to_text_whisper_result())}
-  @type speech_to_text_whisper_result :: %{
+  @type speech_to_text_whisper_output :: %{
+          chunks: list(speech_to_text_whisper_chunk())
+        }
+  @type speech_to_text_whisper_chunk :: %{
           text: String.t(),
-          chunks:
-            list(%{
-              text: String.t(),
-              start_timestamp_seconds: number() | nil,
-              end_timestamp_seconds: number() | nil
-            })
+          start_timestamp_seconds: number() | nil,
+          end_timestamp_seconds: number() | nil
         }
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/bumblebee/audio.ex
```diff
@@ -6,7 +6,27 @@ defmodule Bumblebee.Audio do
   # TODO: remove in v0.5
   @deprecated "Use Bumblebee.Audio.speech_to_text_whisper/5 instead."
   def speech_to_text(model_info, featurizer, tokenizer, generation_config, opts \\ []) do
-    speech_to_text_whisper(model_info, featurizer, tokenizer, generation_config, opts)
+    serving = speech_to_text_whisper(model_info, featurizer, tokenizer, generation_config, opts)
+    client_postprocessing = serving.client_postprocessing
+
+    Nx.Serving.client_postprocessing(serving, fn output_pair, info ->
```
jonatanklosko, line 12: The output format changed, so I'm wrapping to avoid a breaking change.

lib/bumblebee/audio.ex
```diff
@@ -21,15 +41,13 @@ defmodule Bumblebee.Audio do
 
   """
   @type speech_to_text_whisper_input :: Nx.t() | {:file, String.t()}
-  @type speech_to_text_whisper_output :: %{results: list(speech_to_text_whisper_result())}
```
jonatanklosko, line 24: I removed `results: [...]`. It could make sense for non-deterministic text generation with sampling, but for transcription we generally expect a specific output, so I don't think it will ever be useful.

lib/bumblebee/audio/speech_to_text_whisper.ex
```diff
@@ -415,7 +483,7 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
     # above, we would use the second alignment, taking `abc` from the
     # left sequence and `de` from the right one.
 
-    sequences = Enum.map(sequences, &Nx.tensor/1)
+    sequences = Enum.map(sequences, &Nx.tensor(&1, backend: Nx.BinaryBackend))
```
jonatanklosko, line 486:
```suggestion
    # We use binary backend so we are not blocked by the serving computation,
    # in this case we do simple operations with small data so it is fine
    sequences = Enum.map(sequences, &Nx.tensor(&1, backend: Nx.BinaryBackend))
```

jonatanklosko:
@josevalim Nx should be good to go with respect to streaming, awesome job!

lib/bumblebee/audio/speech_to_text_whisper.ex
```diff
@@ -239,11 +262,35 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
...
+    batch_size = length(outputs)
+
+    {lengths, lengths_rest} = Enum.split(state.lengths, batch_size)
+    state = %{state | lengths: lengths_rest}
+
     acc =
-      Enum.reduce(chunk_outputs, acc, fn {sequence, lengths}, acc ->
+      outputs
+      |> Enum.zip(lengths)
+      |> Enum.reduce(state.acc, fn {sequence, lengths}, acc ->
```
josevalim, line 289: We have zip_reduce :)

lib/bumblebee/audio/speech_to_text_whisper.ex
josevalim: I am wondering if it is worth actually creating a tensor, you know. We could keep this as a list, which would help avoid the Nx.to_flat_list at the end as well.

lib/bumblebee/audio.ex
```diff
@@ -21,15 +41,13 @@ defmodule Bumblebee.Audio do
...
+  @type speech_to_text_whisper_chunk :: %{
           text: String.t(),
-          chunks:
-            list(%{
-              text: String.t(),
-              start_timestamp_seconds: number() | nil,
-              end_timestamp_seconds: number() | nil
-            })
+          start_timestamp_seconds: number() | nil,
+          end_timestamp_seconds: number() | nil
```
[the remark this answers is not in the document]
jonatanklosko: Ah you are right it still depends on the option just in a different way. I'm not really a fan of `-1`, though it's the most straightforward option. Typing serving may be tricky on its own because what we call is `Nx.Serving.batched_run` and the return value depends on the serving name we pass?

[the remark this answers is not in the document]
josevalim: Yeah, there is still a lot for it to happen, but I thought I would start the discussion. If we keep this as is, then -1 is like the only viable option in the future if we want to keep a single type.

Another option is to have a `speech_to_text_whisper` function, which returns a string (as it did before) and have `transcribe_whisper` (which has the return type above without nils).

lib/bumblebee/audio/speech_to_text_whisper.ex
jonatanklosko: We could do the comparison with `zip_reduce`, the main difference is slicing, which is more efficient with binary than lists. Though the lists shouldn't be particularly long and in this case the difference is negligible in the context of running the model.

josevalim (approved):

lib/bumblebee/audio.ex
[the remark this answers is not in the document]
jonatanklosko: Let's keep `nil` and if we get to type serving outputs we can switch to `-1`. The issue with separate functions is that if we have another option like that then we have 2x2 functions :D

PR #243, opened 2023-09-13
Move featurizer batch part to serving computation

jonatanklosko:
🐈

### DIFF
--- a/lib/bumblebee/featurizer.ex
+++ b/lib/bumblebee/featurizer.ex
@@ -11,7 +11,69 @@ defmodule Bumblebee.Featurizer do
   @type t :: Bumblebee.Configurable.t()
 
   @doc """
-  Performs feature extraction on the given input.
+  Converts the given input to a batched tensor (or a tensor container).
+
+  Numerical batch processing should be moved to `c:process_batch/2`
+  whenever possible.
   """
-  @callback apply(t(), input :: any(), defn_options :: keyword()) :: any()
+  @callback process_input(t(), input :: any()) :: Nx.t() | Nx.Container.t()
+
+  @doc """
+  Returns an input template for `c:process_batch/2`.
+
+  The shape is effectively the same as the result of `c:process_input/2`,
+  except for the batch size.
+  """
+  @callback batch_template(t(), batch_size :: pos_integer()) :: Nx.t() | Nx.Container.t()
+
+  @doc """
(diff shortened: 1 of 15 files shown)


lib/bumblebee/featurizer.ex
```diff
@@ -11,7 +11,69 @@ defmodule Bumblebee.Featurizer do
...
+  @doc """
+  Optional batch processing stage.
+
+  This is a numerical function. It receives the result of `c:process_input/2`,
+  except the batch size may differ.
+
+  If the featurizer does not define batch processing, the input is
+  returned as is.
   """
-  @callback apply(t(), input :: any(), defn_options :: keyword()) :: any()
```
jonatanklosko, line 16: I didn't deprecate, because it's very unlikely that someone implements a featurizer outside bumblebee.

josevalim: Beautiful. If you want to keep backwards compatibility, you could keep it as a apply and introduce apply_batch.

lib/bumblebee/featurizer.ex
```diff
@@ -11,7 +11,69 @@ defmodule Bumblebee.Featurizer do
...
+  @spec process_batch(t(), Nx.t() | Nx.Container.t()) :: Nx.t() | Nx.Container.t()
+  def process_batch(%module{} = featurizer, batch) do
+    if function_exported?(module, :process_batch, 2) do
```
josevalim, line 73: Keep in mind this will only work if `module` is already loaded (which will be the case if you have called a previous function in module).

josevalim (approved):
Great!

lib/bumblebee/featurizer.ex
jonatanklosko: Ah right, will add ensure loaded!

lib/bumblebee/featurizer.ex
jonatanklosko: I would rather check for it as fallback, I just don't think it's worth in this case.

As for naming, I didn't go with `apply_batch` because then it seems as if `apply_batch` were batched version of `apply`. I'm not sure the current naming is perfect either, but the best I come up with :)

josevalim: Your call, just mentioning for completeness.

PR #252, opened 2023-09-21
Align batch handling for serving run and batched run

jonatanklosko:
Uses the new `Nx.Serving.batch_size/2` and also converts the input to a basic stream.


lib/bumblebee/audio/speech_to_text_whisper.ex
```diff
@@ -104,8 +104,20 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
...
-      inputs = Bumblebee.Featurizer.process_input(featurizer, all_chunks)
-      {Nx.Batch.concatenate([inputs]), {multi?, all_num_chunks, lengths}}
+      if batch_size do
+        stream =
+          all_chunks
+          |> Stream.chunk_every(batch_size)
+          |> Stream.map(fn all_chunks ->
+            inputs = Bumblebee.Featurizer.process_input(featurizer, all_chunks)
+            Nx.Batch.concatenate([inputs])
+          end)
```
jonatanklosko, line 114: @josevalim for now we still read the whole audio, but we chunk, so that the `Bumblebee.Featurizer.process_input` is called per batch, rather than for all chunks at once. I will later look at optimising ffmpeg to only decode chunk at a time.

mix.exs
```diff
@@ -32,12 +32,12 @@ defmodule Bumblebee.MixProject do
...
-      {:torchx, "~> 0.6.1", only: [:dev, :test]},
-      # {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
-      # {:exla, github: "elixir-nx/nx", sparse: "exla", override: true, only: [:dev, :test]},
-      # {:torchx, github: "elixir-nx/nx", sparse: "torchx", override: true, only: [:dev, :test]},
+      # {:nx, "~> 0.6.1"},
+      # {:exla, "~> 0.6.1", only: [:dev, :test]},
+      # {:torchx, "~> 0.6.1", only: [:dev, :test]},
+      {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
+      {:exla, github: "elixir-nx/nx", sparse: "exla", override: true, only: [:dev, :test]},
+      {:torchx, github: "elixir-nx/nx", sparse: "torchx", override: true, only: [:dev, :test]},
```
josevalim, line 40:
```suggestion
      {:nx, "~> 0.6.2"},
      {:exla, ">= 0.0.0", only: [:dev, :test]},
      {:torchx, ">= 0.0.0", only: [:dev, :test]},
      # {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
      # {:exla, github: "elixir-nx/nx", sparse: "exla", override: true, only: [:dev, :test]},
      # {:torchx, github: "elixir-nx/nx", sparse: "torchx", override: true, only: [:dev, :test]},
```

PR #257, opened 2023-09-27
Add link to repository inspector

### LINKED ISSUE #248: Give users a way to tell if Bumblebee supports a model
It is hard for beginners to tell what kind of model bumblebee could support.

I propose a feature that lets you paste the URL of a huggingface model and get a response telling you if bumblebee supports the model or not.

Here is an example of where this could be useful.

Looking at ./lib/bumblebee.ex I see that LLamaForCausalLM is a supported model architecture. But does that mean that I can run codellama?

I can't tell unless I try. Which isn't something most people would be willing to do. Having a script that tells you if a model is supported would be a huge help for beginners.

jonatanklosko:
Closes #248.

image
image

Note that this will benefit from the new error messages after next release :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,9 @@ Here is a list of files commonly found in a repository following the Transformer
 
 ### Model support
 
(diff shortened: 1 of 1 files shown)


a-alhusaini:
I'm going off of OCR here (can't see the image)

This seems to be exactly what we need. I wonder if we could do this without having to download the model.... Would make things even faster :)

jonatanklosko:
> I wonder if we could do this without having to download the model.... Would make things even faster :)

Correct! We are not loading the model parameters, just the model configuration, which is enough to tell if it's supported or not. huggingface/transformers primarily use PyTorch, so basically all repos have PyTorch parameters, which we support, so there's a very low chance that loading the parameters fails.

PR #271, opened 2023-11-02
Add more docs on model caching

ityonemo:


josevalim (approved):

lib/bumblebee.ex
```diff
@@ -1000,6 +1012,9 @@ defmodule Bumblebee do
       {:ok, scheduler} =
         Bumblebee.load_scheduler({:hf, "CompVis/stable-diffusion-v1-4", subdir: "scheduler"})
 
+  ## See also:
+
+  `cache_dir/0` for where the loaded modules are saved
```
jonatanklosko, line 1017: I think it's fine to only do this for `load_model`, because those are the only large files, in case someone want's to clear the cache :)

lib/bumblebee.ex
```diff
@@ -1147,6 +1150,9 @@ defmodule Bumblebee do
 
   @doc """
   Returns the directory where downloaded files are stored.
+
+  Defaults to the standard cache location for the given operating system.
+  Can be overridden with the `BUMBLEBEE_CACHE_DIR` environment variable.
```
jonatanklosko, line 1155:
```suggestion
  Can be configured with the `BUMBLEBEE_CACHE_DIR` environment variable.
```

lib/bumblebee.ex
ityonemo: oh, I didn't fully understand how the Bumblebee cache works.  I assumed it was everything.  Now I understand that it's only the models.  We should probably allow the user to cache everything locally or their own trusted store, because calling out to huggingface is not nice to the huggingface servers and also hf might be down at any given point.

jonatanklosko: Nono, we do cache all files. My point is that `cache_dir` is most useful if you want to free some disk space and that's relevant only for models. All other files are very small, so that's negligible.

jonatanklosko: We actually have some notes on caching at build time and not talking to HF at runtime [here](https://github.com/elixir-nx/bumblebee/tree/main/examples/phoenix#2-cached-from-hugging-face). Also, people can clone HF repos and use `{:local, "/path/to/dir"}` :)

PR #281, opened 2023-11-13
[in-progress] Add credo to project for better code-readability and performance fixes

imahmedismail:
This PR includes the following major changes:

- Add credo 1.7 to elixir-nx/bumblebee
- Fix the code-readability issues in the project:
     - Numbers larger than 9999 should be written with underscores: 38_576
     - Do not use parentheses when defining a function which has no arguments.
     - The alias  is not alphabetically ordered among its group.
     - Skip Credo.Check.Readability.PreferImplicitTry to fix Prefer using an implicit 'try' rather than explicit 'try'.


imahmedismail:
THIS PR IS STILL IN PROGRESS Still need to resolve these issues pointed-out by credo:

image

josevalim:
We have no plan to adopt Credo or others linters in Nx projects. Thank you!

imahmedismail:
Okay thanks for quick-response @josevalim

PR #298, opened 2023-12-06
Apply cross attention spec unet

robinmonjo:
Warning beginner 😋.

I have been trying to use the `UNet2DConditional` model but without cross attention (so basically a `UNet2D` model). 

In python, hugging face have both: [UNet2DCondition](https://huggingface.co/docs/diffusers/api/models/unet2d-cond) and [UNet2D](https://huggingface.co/docs/diffusers/api/models/unet2d).

I need to use an equivalent of `UNet2D`. Seems that it's supported by bumblebee for example in the unet layer in `down_block_2d` there is this doc: 

> When `encoder_hidden_state` is not `nil`, applies cross-attention.

However, from the `UNet2DConditional` model we have no way to have a nil `encoder_hidden_state`. So this PR add an option the model.

I'm not sure this won't bring confusion because from my understanding, without `encoder_hidden_state` this is not a conditional unet anymore.

Let me know what you think.

### DIFF
--- a/lib/bumblebee/diffusion/unet_2d_conditional.ex
+++ b/lib/bumblebee/diffusion/unet_2d_conditional.ex
@@ -72,9 +72,13 @@ defmodule Bumblebee.Diffusion.UNet2DConditional do
       doc:
         "the number of attention heads for each attention layer. Optionally can be a list with one number per block"
     ],
+    apply_cross_attention: [
+      default: true,
+      doc: "weither or not to apply cross attention"
+    ],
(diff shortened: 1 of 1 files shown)


robinmonjo:
Hey @jonatanklosko thank for your feedback. I can't access the link in "this checkpoint", got a 404, I'm probably not allowed to see it 🤷🏻‍♂️

jonatanklosko:
I'm going to close this one. If you want to add the separate model, feel free to open a new PR :)

PR #307, opened 2023-12-14
Apply tokenizer truncation before post-processor

### LINKED ISSUE #306: Token Truncation differs from Transformers implementation
While using the `distilbert-base-uncased` to tokenize a few sentences, I've noticed that the tokens, and, as a result, any classification done differ between the Bumblebee versus the Python Transformers library. 

For example, on Elixir:

```elixir
Mix.install([
  {:nx, "~> 0.5"},
  {:bumblebee, "~> 0.4.2"}
])

{:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "distilbert-base-uncased"})

text = "This is a very long text."
Bumblebee.apply_tokenizer(tokenizer, text,
            length: 10,
            return_token_type_ids: false
          )

%{
  "attention_mask" => #Nx.Tensor<
    u32[1][10]
    [
      [1, 1, 1, 1, 1, 1, 1, 1, 1, 0]
    ]
  >,
  "input_ids" => #Nx.Tensor<
    u32[1][10]
    [
      [101, 2023, 2003, 1037, 2200, 2146, 3793, 1012, 102, 0]
    ]
  >
}

Bumblebee.apply_tokenizer(tokenizer, text,
            length: 5,
            return_token_type_ids: false
          )

%{
  "attention_mask" => #Nx.Tensor<
    u32[1][5]
    [
      [1, 1, 1, 1, 1]
    ]
  >,
  "input_ids" => #Nx.Tensor<
    u32[1][5]
    [
      [101, 2023, 2003, 1037, 2200]
    ]
  >
}

```

We can see that when truncating, Bumblebee doesn't add the `[SEP]` (102) token at the end. 

While on Transformers:

```python
import torch
from transformers import DistilBertTokenizerFast, DistilBertForSequenceClassification

tokenizer = DistilBertTokenizerFast.from_pretrained("distilbert-base-uncased")

device = torch.device('cpu')
text = 'This is a very long text.'

tokens = tokenizer(text, truncation=True, padding=True, return_tensors="pt", max_length=10).to(device)
{'input_ids': tensor([[ 101, 2023, 2003, 1037, 2200, 2146, 3793, 1012,  102]]), 'attention_mask': tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1]])}

tokens = tokenizer(text, truncation=True, padding=True, return_tensors="pt", max_length=5).to(device)
{'input_ids': tensor([[ 101, 2023, 2003, 1037,  102]]), 'attention_mask': tensor([[1, 1, 1, 1, 1]])}
```

Transformers always has the token 102 at the end, even when truncating. I've noticed this behavior because the model trained using Transformers was not giving the same results when run by `Bumblebee.Text.text_classification`. After a bit of digging, I found out that it only happened when the token string was truncated.

jonatanklosko:
Closes #306.

### DIFF
--- a/lib/bumblebee/utils/tokenizers.ex
+++ b/lib/bumblebee/utils/tokenizers.ex
@@ -21,14 +21,26 @@ defmodule Bumblebee.Utils.Tokenizers do
 
     input = List.wrap(input)
 
+    tokenizer =
+      if length = opts[:length] do
+        upper_bound_length = length |> List.wrap() |> Enum.max()
+
+        Tokenizer.set_truncation(tokenizer,
+          max_length: upper_bound_length,
+          direction: opts[:truncate_direction]
+        )
+      else
+        tokenizer
+      end
+
     {:ok, encodings} =
       Tokenizer.encode_batch(tokenizer, input, add_special_tokens: opts[:add_special_tokens])
 
     length = opts[:length]
 
-    {pad_length, truncate_length} =
+    pad_length =
       if is_number(length) do
-        {length, length}
+        length
       else
         max_length =
           encodings
@@ -36,12 +48,8 @@ defmodule Bumblebee.Utils.Tokenizers do
           |> Enum.max()
 
         case length do
-          nil ->
-            {max_length, nil}
-
-          lengths when is_list(lengths) ->
-            bounding_length = find_bounding_length(max_length, lengths)
-            {bounding_length, bounding_length}
+          nil -> max_length
+          lengths when is_list(lengths) -> find_bounding_length(max_length, lengths)
         end
       end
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):

lib/bumblebee/utils/tokenizers.ex
```diff
@@ -21,55 +21,47 @@ defmodule Bumblebee.Utils.Tokenizers do
...
     input = List.wrap(input)
 
+    tokenizer =
+      if length = opts[:length] do
+        upper_bound_length = length |> List.wrap() |> Enum.max()
+
+        Tokenizer.set_truncation(tokenizer,
+          max_length: upper_bound_length,
+          direction: opts[:truncate_direction]
+        )
```
jonatanklosko, line 31: When we apply the tokenizer it does:

1. Truncate the sequence if configured (accounting for the number of tokens added by 2.)
2. Applies post-processor (in particular templating that adds leading trailing tokens)
3. Applies padding if configured

So for example, lets say we have sequence that tokenizes to [10, 20, 30, 40] and the processor that adds 0 at the beginning and 1 at the end. And let's say we want to enforce length 4. If we tokenize without truncation, it's going to give us [0, 10, 20, 30, 40, 1], and then we manually truncate to [0, 10, 20, 30], which looses the added end token. On the other hand if we configure truncation it's going to compute [10, 20, 30, 40], then first truncate to 4 - 2 addeditional tokens, so [10, 20], and then add the tokens [0, 10, 20, 1].

---

Now the issue is that `Tokenizer.set_truncation` copies the underlying tokenizer. It is relatively cheap (bumping some refcounts and copying `added_vocabulary`, not the full vocabulary). That said ideally we would configure the tokenizer once. So what we can do deprecate options like `:length`, `:truncate_direction`, `:pad_direction`, and instead have a separate function to configure that. @josevalim wdyt?

josevalim: As you prefer!

PR #311, opened 2023-12-15
Add :type option to load model under specific precision

jonatanklosko:
Allows for `Bumblebee.load_model({:hf, "..."}, type: :bf16)` to set mixed precision policy on the model and cast params on load. The user can also pass the policy struct itself, type is just a shorthand.

Note: currently loading a f16 checkpoint returns params in f16, but that's somewhat a bug, specifically if all params are available we skip Axon init, which would otherwise cast the params to f32. Now it's going to always be f32 by default and the user can override it using `:type`.

hf/transformers look at `"torch_dtype"` in the config file to determine the type, but I'm not sure if we want to configure the policy automatically based on that. @seanmor5 thoughts?

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -317,6 +317,11 @@ defmodule Bumblebee do
   @doc """
   Builds an `Axon` model according to the given specification.
 
+  ## Options
+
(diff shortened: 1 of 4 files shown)


josevalim (approved):
LGTM!

PR #319, opened 2023-12-19
Remove Bumblebee.Utils.Nx.to_list/1 in favour of Nx.to_list/1

thiagopromano:
After `Nx.to_list/1` (https://github.com/elixir-nx/nx/pull/1095) was added, the function `Bumblebee.Utils.Nx.to_list` became redundant.

### DIFF
--- a/lib/bumblebee/utils/nx.ex
+++ b/lib/bumblebee/utils/nx.ex
@@ -64,26 +64,6 @@ defmodule Bumblebee.Utils.Nx do
   end
 
   @doc """
-  Returns the underlying tensor as a list.
-
-  The list nesting matches the tensor shape.
-  """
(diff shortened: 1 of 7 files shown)


lib/bumblebee/utils/nx.ex
```diff
@@ -68,6 +68,8 @@ defmodule Bumblebee.Utils.Nx do
 
   The list nesting matches the tensor shape.
   """
+  # TODO: remove in v0.6
+  @deprecated "Use Nx.to_list/1 instead."
```
jonatanklosko, line 72: This is a private module, you can totally just remove the function :D

jonatanklosko (approved):
Thanks!

PR #322, opened 2023-12-22
Pass PRNG key to schedulers in Stable Diffusion

jonatanklosko:
Propagates the per-input seed all the way to the scheduler.

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -1048,10 +1048,11 @@ defmodule Bumblebee do
   @spec scheduler_init(
           Bumblebee.Scheduler.t(),
           non_neg_integer(),
-          tuple()
+          Nx.Tensor.t(),
+          Nx.Tensor.t()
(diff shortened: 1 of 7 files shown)


lib/bumblebee.ex
```diff
@@ -1048,10 +1048,11 @@ defmodule Bumblebee do
   @spec scheduler_init(
           Bumblebee.Scheduler.t(),
           non_neg_integer(),
-          tuple()
+          Nx.Tensor.t(),
+          Nx.Tensor.t()
         ) :: {Bumblebee.Scheduler.state(), Nx.Tensor.t()}
-  def scheduler_init(%module{} = scheduler, num_steps, sample_shape) do
-    module.init(scheduler, num_steps, sample_shape)
+  def scheduler_init(%module{} = scheduler, num_steps, sample_template, prng_key) do
```
jonatanklosko, line 1054: Could also be an option, but since we apply vectorization both `sample_template` and `prng_key` I think I like it more as an argument.

PR #326, opened 2024-01-28
Add zero-shot vision

seanmor5:
Needs tests


jonatanklosko:
Looks good, we just need tests and docs with example :)

PR #328, opened 2024-02-08
Less restrictive unzip dependency

sonic182:
Can we have a less restrictive unzip dependency? It breaks if the project that uses bumblebee, uses a newer unzip.


jonatanklosko:
I updated the requirement to `~> 0.10.0`. We can't relax it more, because then we may run into breaking changes. Note that the user can still use a newer version with `override: true`, they just need to verify that it still works as expected (that is, the APIs Bumblebee uses didn't change).

PR #332, opened 2024-02-13
Group cache files by repository

jonatanklosko:
Currently our cache directory has flat structure, where for each cached file we have one `.json` file with metadata and another with file contents. The file names are URL hashes or etags, so they are not meaningful to the user.

The cache should be opaque to the user, however it's easy to accumulate a lot of models, but clearing the whole cache is not ideal, since the user may want to remove only the models they no longer plan to use. For this purpose, this PR groups all cache files related to a specific HF repository in a separate subdirectory with meaningful name.

As a result, the existing caches will be invalidated, but I think it's worth it and we can just mention in the changelog that users should clear the cache.

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -1163,11 +1163,12 @@ defmodule Bumblebee do
   defp get_repo_files({:hf, repository_id, opts}) do
     subdir = opts[:subdir]
     url = HuggingFace.Hub.file_listing_url(repository_id, subdir, opts[:revision])
+    cache_scope = repository_id_to_cache_scope(repository_id)
 
     result =
       HuggingFace.Hub.cached_download(
         url,
-        Keyword.take(opts, [:cache_dir, :offline, :auth_token])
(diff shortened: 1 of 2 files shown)


lib/bumblebee.ex
```diff
@@ -1211,13 +1212,21 @@ defmodule Bumblebee do
...
-      [etag: etag] ++ Keyword.take(opts, [:cache_dir, :offline, :auth_token])
+      [etag: etag, cache_scope: cache_scope] ++
+        Keyword.take(opts, [:cache_dir, :offline, :auth_token])
     )
   end
 
+  defp repository_id_to_cache_scope(repository_id) do
+    repository_id
+    |> String.replace("/", "--")
+    |> String.replace(~r/[^\w-]/, "")
```
josevalim, line 1227: Maybe we can Base.url_encode64 it instead?

jonatanklosko: The idea is exactly to keep it readable, so we end up with subdirs like `facebook--convnext-tiny-224` :)

PR #350, opened 2024-02-25
add support for Nomic embed model v1 & 1.5

nyo16:
The Nomic model for embeds looks promising and supporting 8k seq. 
[HG link](https://huggingface.co/nomic-ai/nomic-embed-text-v1.5)


jonatanklosko:
Hey @nyo16, the source code for Nomic is currently on HF Hub ([here](https://huggingface.co/nomic-ai/nomic-embed-text-v1-unsupervised/blob/main/modeling_hf_nomic_bert.py)), which means it diverges from Bert in some details. On a quick glance it uses rotary embedding scaling (to support longer token input), which is not available in Bert. Code on the Hub is used for new models and it can change if the model is still being improved. Popular models are eventually added to hf/transformers, once they stabilise. Once it is added to hf/transformers we can add it as well :)

PR #359, opened 2024-03-05
Add Stable Diffusion ControlNet

joelpaulkoch:
I want to share my work on using ControlNet with Stable Diffusion. These are the three parts and notes on current limitations:

1. ControlNet model
- there is a conditioning scale parameter in the `diffusers` implementation, at the moment I have a constant of 1 for this. Would you add it as (optional) input?
- regarding testing: there is "hf-internal-testing/tiny-controlnet", but this only returned zeros for me (in `diffusers` and `bumblebee`), so I kept the test with "lllyasviel/sd-controlnet-scribble".

2. UNet
- I added a new model architecture `:with_additional_residuals` and separate input and core functions. But in the end the only difference is that additional residuals are passed in and added. So alternatively this could go in the `:base` architecture as optional inputs and add layers I guess.

3. Stable Diffusion with ControlNet
- Similarly, I've copied the existing stable diffusion implementation and adapted it to support the control net. It might be better to have it in the existing StableDiffusion module.
- the current implementation accepts a u8 tensor of the correct size as conditioning image. Preprocessing is converting the tensor to f32. It might make sense to be more lenient e.g. resize the conditioning image as part of the preprocessing (?)

I've tried all the control nets listed [here](https://huggingface.co/lllyasviel/sd-controlnet-scribble#released-checkpoints) with the corresponding example and got sensible results for all but the [normal map one](https://huggingface.co/lllyasviel/sd-controlnet-normal). I'm not sure what's the issue with the normal map, but could imagine it's because of the preprocessing, or I simply did not run enough steps.


jonatanklosko:
Hey @joelpaulkoch, thanks for the PR! I will have a more detailed look later, for now a couple high-level comments :)

> there is a conditioning scale parameter in the diffusers implementation, at the moment I have a constant of 1 for this.

Having is a optional serving input sounds good (similar to how we have :seed, for example).

> I added a new model architecture :with_additional_residuals and separate input and core functions.

Since the difference is only in inputs, I would totally just have them as optional inputs in the `:base` architecture, yeah. This also more closely matches what hf/transformers do.

> Similarly, I've copied the existing stable diffusion implementation and adapted it to support the control net. It might be better to have it in the existing StableDiffusion module.

I think a separate module makes sense, in general I would have one module per diffusion type (SD, SD control net, SD XL, ...) and then a serving function for a task (currently only `text_to_image`, but could be `image_to_image`, and so on). This would roughly correspond to the diffusers, such serving function corresponds to pipeline class and module is the pipeline directory they have.

> Preprocessing is converting the tensor to f32.

I see in diffusers they have `VaeImageProcessor`, though if in this case it always comes down to converting into f32, then it's probably fine to just be a function.

lib/bumblebee/diffusion/stable_diffusion/control_net.ex
```diff
@@ -0,0 +1,505 @@
...
+
+    sample =
+      for {{in_channels, out_channels}, i} <- Enum.with_index(channels),
+          reduce: state do
+        input ->
+          input
+          |> Axon.conv(in_channels,
+            kernel_size: 3,
+            padding: [{1, 1}, {1, 1}],
+            name: name |> join(2 * i) |> join("conv"),
```
jonatanklosko, line 305:
```suggestion
            name: name |> join("inner_convs") |> join(2 * i),
```

lib/bumblebee/diffusion/stable_diffusion/control_net.ex
```diff
@@ -0,0 +1,505 @@
...
+            activation: :silu
+          )
+          |> Axon.conv(out_channels,
+            kernel_size: 3,
+            padding: [{1, 1}, {1, 1}],
+            strides: 2,
+            name: name |> join(2 * i + 1) |> join("conv"),
```
jonatanklosko, line 312:
```suggestion
            name: name |> join("inner_convs") |> join(2 * i + 1),
```

lib/bumblebee/diffusion/stable_diffusion/control_net.ex
```diff
@@ -0,0 +1,505 @@
...
+
+      blocks_mapping =
+        ["down_blocks.{n}", "mid_block"]
+        |> Enum.map(&Transformers.Utils.prefix_params_mapping(block_mapping, &1, &1))
+        |> Enum.reduce(&Map.merge/2)
+
+      controlnet = %{
+        "controlnet_down_blocks.{m}.zero_conv" => "controlnet_down_blocks.{m}",
+        "controlnet_cond_embedding.input_conv" => "controlnet_cond_embedding.conv_in",
+        "controlnet_cond_embedding.{m}.conv" => "controlnet_cond_embedding.blocks.{m}",
```
jonatanklosko, line 491:
```suggestion
        "controlnet_cond_embedding.inner_convs.{m}" => "controlnet_cond_embedding.blocks.{m}",
```

jonatanklosko (approved):
Thanks for the updates, some more comments :)

lib/bumblebee/diffusion/stable_diffusion_controlnet.ex
```diff
@@ -0,0 +1,534 @@
...
+      prompt = "numbat in forest, detailed, digital art"
+
+      conditioning =
+        Nx.tensor(
+          [for(_ <- 1..8, do: [255]) ++ for(_ <- 1..24, do: [0])],
+          type: :u8
+        )
+        |> Nx.tile([256, 8, 3])
+        |> Nx.pad(0, [{192, 64, 0}, {192, 64, 0}, {0, 0, 0}])
+        |> Nx.transpose(axes: [1, 0, 2])
```
joelpaulkoch, line 117: This is an example that controls the stripes of the numbat. You might want to rewrite this

lib/bumblebee/diffusion/stable_diffusion/control_net.ex
joelpaulkoch: This actually results in a `MatchError` in `Bumblebee.Conversion.PyTorch.match_template/3`. I remember that I encountered this error before. I believe something doesn't work out with the prefixes and I "fixed" it by not ending the key with a substitution.

jonatanklosko: Ah right, I've just fixed it :)

lib/bumblebee/diffusion/stable_diffusion_controlnet.ex
jonatanklosko: Could be a nice exercise to add a notebook that detects edges with Evision and uses it as edges control ([like here](https://huggingface.co/blog/controlnet)), for the example here any tensor is fine though, just to give an idea. I also added a comment as a clarification :)

jonatanklosko:
We could share more logic between the servings, but it's fine for now, we can refactor once there are more :)

jonatanklosko (approved):
@joelpaulkoch it's good to go, thanks for working on this!

PR #361, opened 2024-03-11
Support stream input in Whisper serving and stream ffmpeg chunks

### LINKED ISSUE #261: Stream audio chunk by chunk to Whisper
Hey, it's already possible to provide data to the model in chunks, however, it seems that the whole audio has to be available at once, which is impossible for live streaming. Would it be possible to support streaming the audio to the serving chunk by chunk?

jonatanklosko:
Closes #261.

This allows the Whisper serving to accept a stream of consecutive chunks. Importantly, this improves the `{:file, path}`, such that we read the file in chunks using ffmpeg, rather than loading it into memory all at once.

This PR drops support for a list if inputs, such as `Nx.Serving.batched_run(MyServing, [{:file, path1}, {:file2, path2}]`). This serving works on a higher level than usually, because a single chunked input is already multiple inputs to the model, so I think this is sane. Multiple inputs can be processed concurrently by calling `batched_run` from multiple processes.

I noticed a bug, specifically when streaming with timestamps and using a `batch_size`, we would ignore small segments after every batch.

---

@josevalim I went with the stream of consecutive chunks and handle accumulation + overlapping internally. I think it is preferable to keep the chunking details internal and I don't see a benefit of exposing it. If we shift accumulation to the user, they would basically need to do exactly that. For ffmpeg we would do that in bumblebee anyway, because accumulation is better than duplicating ffmpeg decoding work.

### DIFF
--- a/lib/bumblebee/audio.ex
+++ b/lib/bumblebee/audio.ex
@@ -10,11 +10,14 @@ defmodule Bumblebee.Audio do
 
     * a 1-dimensional `Nx.Tensor` with audio samples
 
+    * an enumerable of 1-dimensional `Nx.Tensor`s, represending a
+      continuous stream of input
+
     * `{:file, path}` with path to an audio file (note that this
       requires `ffmpeg` installed)
 
   """
-  @type audio :: Nx.t() | {:file, String.t()}
+  @type audio :: Nx.t() | Enumerable.t(Nx.t()) | {:file, String.t()}
 
   @type speech_to_text_whisper_input ::
           audio() | %{:audio => audio(), optional(:seed) => integer() | nil}
@@ -31,7 +34,11 @@ defmodule Bumblebee.Audio do
   Builds serving for speech-to-text generation with Whisper models.
 
   The serving accepts `t:speech_to_text_whisper_input/0` and returns
-  `t:speech_to_text_whisper_output/0`. A list of inputs is also supported.
+  `t:speech_to_text_whisper_output/0`.
+
+  This serving always accepts a single input. A list of tensors is
+  interpreted as continuous chunks. To transcribe multiple inputs
+  concurrently use `Nx.Serving.batched_run/2`.
 
   ## Options
 
--- a/lib/bumblebee/audio/speech_to_text_whisper.ex
+++ b/lib/bumblebee/audio/speech_to_text_whisper.ex
@@ -30,7 +30,10 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
     Shared.validate_architecture!(spec, [:for_conditional_generation])
 
     chunk_num_seconds = opts[:chunk_num_seconds]
-    context_num_seconds = opts[:context_num_seconds]
+
+    context_num_seconds =
+      opts[:context_num_seconds] || (chunk_num_seconds && chunk_num_seconds / 6)
+
     preallocate_params = opts[:preallocate_params]
     defn_options = opts[:defn_options]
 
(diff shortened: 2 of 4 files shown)


lib/bumblebee/audio.ex
```diff
@@ -10,11 +10,14 @@ defmodule Bumblebee.Audio do
 
     * a 1-dimensional `Nx.Tensor` with audio samples
 
+    * an enumerable of 1-dimensional `Nx.Tensor`s, represending a
+      continuous stream of input
```
josevalim, line 14: Do they have to be the of the same size? It may be worth adding a line.

lib/bumblebee/audio/speech_to_text_whisper.ex
```diff
@@ -76,101 +86,144 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
...
+              Integer.to_string(channels),
+              "-ar",
+              Integer.to_string(sampling_rate),
+              "-f",
+              format,
+              "-hide_banner",
+              "-loglevel",
+              "quiet",
+              "pipe:1"
+            ])
```
josevalim, line 211:
```suggestion
            System.cmd(
              "ffmpeg",
              ~w[-ss #{offset} -t #{chunk_size} ... -hide_banner -loglevel quiet pipe:1]
            )
```

The above would also work if you prefer it.

josevalim (approved):
Tiny comments but ship as you prefer!

lib/bumblebee/audio.ex
jonatanklosko: They don't, I will add a note!

lib/bumblebee/audio/speech_to_text_whisper.ex
```diff
@@ -76,101 +86,144 @@ defmodule Bumblebee.Audio.SpeechToTextWhisper do
...
-            Nx.Batch.concatenate([{inputs, seed}])
-          end)
+      stream =
+        chunks
+        # Unless batch_size is specified, we process chunks individually.
+        # When the whole audio chunk is given at once, this is all we
+        # need. When an enumerable is given, it can have an arbitarary
+        # number of elements, so processing them one by one makes for a
+        # reasonable default.
+        |> Stream.chunk_every(batch_size || 1)
```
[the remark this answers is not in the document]
jonatanklosko: > I believe the batches do not need to have the same size, right?

In which case they wouldn't be the same? By this point there are no batches, it's the `chunk_every` that we use to create a batch, before we pass to the model.

But now I think we need a separate option for controlling the batch size each client uses, because it may not necessarily be `batch_size`. When all the input is known upfront (streaming a file), we can aggressively use `batch_size` and that doesn't introduce a delay. But when the input comes live from a microphone, waiting for `batch_size` chunks to be present introduces a delay, so in that case we most likely want the client to run as soon as a single chunk is available.

> Unless all of the chunks need to have the same size?

For clarity, in general by chunk I mean a slice of the input audio and chunk size is the audio length, it is rather orthogonal to batching.

lib/bumblebee/audio/speech_to_text_whisper.ex
jonatanklosko: Pretty!

lib/bumblebee/audio/speech_to_text_whisper.ex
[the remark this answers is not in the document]
jonatanklosko: I added `:client_batch_size`, if you have any concerns let me know!

PR #362, opened 2024-03-12
Gemma fixes

jonatanklosko:
See https://github.com/huggingface/transformers/pull/29402.

### DIFF
--- a/lib/bumblebee/layers.ex
+++ b/lib/bumblebee/layers.ex
@@ -1142,9 +1142,7 @@ defmodule Bumblebee.Layers do
       |> Nx.pow(2)
       |> Nx.mean(axes: [opts[:channel_index]], keep_axes: true)
 
-    x =
-      input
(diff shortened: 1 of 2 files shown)


lib/bumblebee/layers.ex
```diff
@@ -1118,35 +1143,66 @@ defmodule Bumblebee.Layers do
...
+    normalized_input =
+      input
+      |> Nx.as_type(:f32)
+      |> rms_normalize(opts)
+      |> Nx.as_type(Nx.type(input))
+
+    normalized_input * (opts[:shift] + weight)
+  end
+
+  defnp rms_norm_impl_upcast_all(input, weight, opts \\ []) do
```
jonatanklosko, line 1188: Initially I was adding conditionals on the options, but I think it's cleaner to just have two separate versions altogether.

josevalim (approved):

PR #364, opened 2024-03-20
Add Gemma attention head size

cmeon:
This fix adds `attention_head_size` which is different for Gemma, equal to 256.

Without this, during module building, the runtime would ignore the mismatched shapes and result in unintelligible results.

```
00:15:21.191 [debug] the following parameters were ignored, because of non-matching shape:

  * decoder.blocks.24.self_attention.value.kernel (expected {3072, 3072}, got: {3072, 4096})
  * decoder.blocks.18.self_attention.value.kernel (expected {3072, 3072}, got: {3072, 4096})
  * decoder.blocks.18.self_attention.output.kernel (expected {3072, 3072}, got: {4096, 3072})
  ...
```

See:
- https://github.com/google-deepmind/gemma/issues/4

### DIFF
--- a/lib/bumblebee/text/gemma.ex
+++ b/lib/bumblebee/text/gemma.ex
@@ -26,6 +26,10 @@ defmodule Bumblebee.Text.Gemma do
         default: 24576,
         doc: "the dimensionality of intermediate layers"
       ],
+      attention_head_size: [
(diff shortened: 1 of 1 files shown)


jonatanklosko (approved):
Good catch, thanks!

PR #374, opened 2024-05-29
Add Phi-3 model

bowyern:
My attempt at adding phi3-4k to Bumblebee. The outputs aren't right, and in talking with Jonatan it looks like there's a problem between the embedder and where the FFN is applied in the decoder layers.


jonatanklosko (approved):

jonatanklosko:
Thanks!

PR #375, opened 2024-05-30
Use model state

### LINKED ISSUE #372: Using bumblebee from git is broken
Since there where minor improvements on github since v0.5.3 and no new released version I'm using bumblebee from git in my mix.exs like `{:bumblebee, git: "https://github.com/elixir-nx/bumblebee"}`. This seems currently broken and when trying to `.load_model/2` it fails due to some Axon changes which I can not debug where the change might come from:

```elixir
iex(3)> {:ok, model_info} = Bumblebee.load_model(repo, backend: {EXLA.Backend, client: :host})
[warning] passing parameter map to initialization is deprecated, use %Axon.ModelState{} instead
** (UndefinedFunctionError) function Axon.ModelState.fetch/2 is undefined (Axon.ModelState does not implement the Access behaviour

You can use the "struct.field" syntax to access struct fields. You can also use Access.key!/1 to access struct fields dynamically inside get_in/put_in/update_in)
    (axon 0.6.1) Axon.ModelState.fetch(%Axon.ModelState{data: %{....
```

```
Erlang/OTP 26 [erts-14.2.4] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [jit:ns]

Interactive Elixir (1.16.2) - press Ctrl+C to exit (type h() ENTER for help)
```

Any ideas?

seanmor5:
Resolves #372

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -299,7 +299,7 @@ defmodule Bumblebee do
   """
   @type model_info :: %{
           model: Axon.t(),
-          params: map(),
+          params: %Axon.ModelState{},
           spec: Bumblebee.ModelSpec.t()
         }
 
--- a/lib/bumblebee/conversion/pytorch_params.ex
+++ b/lib/bumblebee/conversion/pytorch_params.ex
@@ -169,7 +169,7 @@ defmodule Bumblebee.Conversion.PyTorchParams do
       end)
 
     params_data = Map.new(params)
-    params = %{params_expr | data: params_data} |> IO.inspect()
+    params = %{params_expr | data: params_data}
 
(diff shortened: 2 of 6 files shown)


jonatanklosko:
@seanmor5 `model_state = Axon.trace_init(model, input_template)` this still warns because it expects the struct, should there be a function to build the struct from the map perhaps?

seanmor5:
@jonatanklosko I updated trace_init to use `Axon.ModelState.empty`, and then pretty much everywhere in Axon we now use the `Axon.ModelState.new` to avoid warnings

jonatanklosko (approved):

PR #390, opened 2024-08-08
Fix error handling when writing to file during download

jonatanklosko:
The current clauses were never matching.


lib/bumblebee/utils/http.ex
```diff
@@ -122,6 +116,15 @@ defmodule Bumblebee.Utils.HTTP do
     end
   end
 
+  defp safe_binwrite(file, iodata) do
+    try do
+      IO.binwrite(file, iodata)
+    catch
+      :error, error ->
```
jonatanklosko, line 123: This would fail when the file is suddenly closed, not sure when else. We could probably just assert `:ok`, @josevalim let me know your preference :)

josevalim (approved):

PR #392, opened 2024-08-14
Add M2M100 & NLLB model

aymanosman:
The values used in the tests were obtained by running this python script:

```python
from torch import tensor
from transformers import M2M100Model, M2M100ForConditionalGeneration

base = M2M100Model.from_pretrained("hf-internal-testing/tiny-random-M2M100Model")
generation = M2M100ForConditionalGeneration.from_pretrained("hf-internal-testing/tiny-random-M2M100ForConditionalGeneration")

inputs = {
    'input_ids': tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]]),
    'decoder_input_ids':  tensor([[15, 25, 35, 45, 55, 65, 0, 0]]),
    'attention_mask': tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])
}

base(**inputs).last_hidden_state[:,1:4,1:4]
generation(**inputs).logits[:,1:4,1:4]
```

```python
from transformers import AutoModelForSeq2SeqLM

generation = AutoModelForSeq2SeqLM.from_pretrained("hf-internal-testing/tiny-random-nllb")

inputs = {
    'input_ids': tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]]),
    'decoder_input_ids':  tensor([[15, 25, 35, 45, 55, 65, 0, 0]]),
    'attention_mask': tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])
}

generation(**inputs).logits[:,1:4,1:4]
```

Here is a [notebook](https://gist.github.com/aymanosman/00b525b9a2a430870c247321ddbafb7b) demonstrating that it works, but there are issues regarding adding the language codes.


aymanosman:
I also saw these warnings that I do not understand:

```
21:24:16.049 [debug] the following parameters were missing:

  * language_modeling_head.logits_bias.bias


21:24:16.049 [debug] the following PyTorch parameters were unused:

  * lm_head.weight
  * ```

File: lib/bumblebee/text/m2m100.ex [hunk 1]
```diff
@@ -0,0 +1,516 @@
...
+  end
+
+  def model(%__MODULE__{architecture: :for_conditional_generation} = spec) do
+    inputs = encoder_decoder_inputs(spec)
+    outputs = core(inputs, spec)
+
+    logits =
+      outputs.hidden_state
+      |> language_modeling_head(spec, name: "language_modeling_head")
+      |> Axon.bias(name: "language_modeling_head.logits_bias", bias_initializer: :zeros)
```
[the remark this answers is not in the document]
jonatanklosko: Btw. this is the warning you were getting. It couldn't find parameters for this layer (since it's not present in the original model). It still worked because with `bias_initializer: :zeros` it was just adding zeros.

lib/bumblebee/text/m2m100.ex
```diff
@@ -0,0 +1,516 @@
...
+      )
+
+    if spec.scale_embedding do
+      Axon.nx(input_embeddings, fn x -> Nx.multiply(x, Nx.sqrt(spec.hidden_size)) end)
+    else
+      input_embeddings
+    end
+  end
+
+  defp position_embedding(position_ids, spec, opts) do
```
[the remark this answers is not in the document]
jonatanklosko: I think it's fine to keep it here until we need it elsewhere!

We actually have a similar implementation as part of the transformer RoPE (when enabled enabled), as used by the recent LLMs. But there are slight differences and the implementation here is pretty short, so let's keep that.

lib/bumblebee/text/m2m100.ex
```diff
@@ -0,0 +1,516 @@
...
+
+    offset = 2
+    embedding_dim = spec.hidden_size
+    num_embeddings = spec.max_positions + offset
+    padding_idx = spec.pad_token_id
+    half_dim = div(embedding_dim, 2)
+
+    position_ids
+    |> Axon.nx(
+      fn position_ids ->
```
jonatanklosko, line 345: You can define `defnp position_embedding_impl`, so that you can use operators like `-` and `*`, which should make it more clear.

lib/bumblebee/text/m2m100.ex
```diff
@@ -0,0 +1,516 @@
...
+        doc: "the dropout rate for activations inside fully connected layers"
+      ],
+      initializer_scale: [
+        default: 0.02,
+        doc:
+          "the standard deviation of the normal initializer used for initializing kernel parameters"
+      ]
+    ] ++
+      Shared.common_options([:num_labels, :id_to_label]) ++
+      Shared.token_options(pad_token_id: 1, eos_token_id: 2, decoder_start_token_id: 2)
```
jonatanklosko, line 78:
```suggestion
    ] ++ Shared.token_options(pad_token_id: 1, eos_token_id: 2, decoder_start_token_id: 2)
```

(we don't have classification architecture here, so these are not relevant)

jonatanklosko:
@aymanosman thanks for the PR, looks clean, and great job adding the tests :)

lib/bumblebee/text/m2m100.ex
aymanosman: I've given the anonymous function its own `defnp`. Let me know if you think it is more readable.

jonatanklosko: The PyTorch implementation computes embeddings for all possible positions upfront and  then on inference takes the relevant tensors. We generally don't do that, so instead of computing for all positions and doing `Nx.take`, in this case we can bake the positions into the computation directly. I simplified the implementation and used slightly different operations to match what we do in RoPE, it is an equivalent expression, just after some mathematical transformations.

jonatanklosko (approved):
@aymanosman thanks! I will have a look at improving the language configuration soon :)

jonatanklosko:
@aymanosman here's how you can specify the source and target language as expected by the model:

```elixir
source_lang = "eng_Latn"
target_lang = "pol_Latn"

target_token = Bumblebee.Tokenizer.token_to_id(tokenizer, target_lang)
generation_config = Bumblebee.configure(generation_config, forced_bos_token_id: target_token)

serving = Bumblebee.Text.generation(model_info, tokenizer, generation_config)

text = "The bank of the river is beautiful in spring"
Nx.Serving.run(serving, "#{source_lang} #{text}")
```

(From brief testing it seems that specifying the source language improves the translation quality)

I am still considering how to best fit this into Bumblebee APIs, but this should unblock your testing :)

aymanosman:
@jonatanklosko Thanks. Looking forward to the improved API. I was thinking of modifying the generation serving to take structured input that contained more than just the input text but also the source and target languages.

So something like this would be possible:

```elixir
Nx.Serving.run(serving, {"eng_Latn", "fra_Latn", "The bank of the river is beautiful in spring"})
```

But I'll wait for you to come up with a better idea :).

jonatanklosko:
@aymanosman implemented in #395 :)

PR #393, opened 2024-08-15
Fix typo in load_tokenizer's ArgumentError

chaychoong:
A prior commit 8b52612 removed the concept of tokenizer modules and introduced tokenizer types. As part of the change, the sole expected option for the `load_tokenizer` function was changed from `:module` to `:type`.

Unfortunately the commit missed out on changing the error message, which could cause slight confusion amongst devs that use `load_tokenizer` when upgrading bumblebee from pre-v0.5.0.

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -911,7 +911,7 @@ defmodule Bumblebee do
                   type
 
                 {:error, error} ->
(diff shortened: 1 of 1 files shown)


jonatanklosko (approved):
Thanks!

PR #394, opened 2024-08-19
Add Swin model

bosko:


jonatanklosko:
Thanks! I fixed the issues we discussed and did some refactoring, it works as expected now :)

PR #400, opened 2024-10-07
Rewrite phoenix examples with phoenix_playground

### LINKED ISSUE #396: Support cache: binary()
EXLA introduced a `cache: binary()` option to improve boot times. However, that option requires a file path and, if a single Bumblebee serving compiles multiple models (such as StableDiffusion or batch_keys), the caches will end-up overlapping each other. We should probably append a suffix to the `:cache` key if it is a binary. For example, we should probably require an atom argument as first argument on `compile_or_jit`:

https://github.com/elixir-nx/bumblebee/blob/9421ecad97077a530626955840ccb7d5022fa7b7/lib/bumblebee/shared.ex#L325

We should probably update the Phoenix examples to use this option too.

jonatanklosko:
Also adds `:cache` to the examples. Closes #396.

cc @wojtekmach :D

### DIFF
--- a/examples/phoenix/image_classification.exs
+++ b/examples/phoenix/image_classification.exs
@@ -1,17 +1,5 @@
-Application.put_env(:sample, PhoenixDemo.Endpoint,
-  http: [ip: {127, 0, 0, 1}, port: 8080],
-  server: true,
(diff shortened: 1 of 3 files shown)


wojtekmach:
image

🤝

PR #402, opened 2024-10-17
Update Mistral LLMs notebook

caioaao:
- Add instructions to request access to the model (now required for Mistral)
- Reduce `sequence_length` so notebook is more accessible - I got OOM even though I have a 32Gb RAM laptop


jonatanklosko (approved):
Thanks :)

PR #404, opened 2024-10-30
Add Phoenix examples readme to docs

jonatanklosko:
The only issue I see is that the URLs point to main rather than docs-version branch. @josevalim let me know if you prefer a different approach.

### DIFF
--- a/examples/phoenix/README.md
+++ b/examples/phoenix/README.md
@@ -1,6 +1,6 @@
 # Phoenix LiveView examples
 
-The [examples/phoenix](https://github.com/elixir-nx/bumblebee/tree/main/examples/phoenix) directory contains contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
+The [examples/phoenix](https://github.com/elixir-nx/bumblebee/tree/main/examples/phoenix) directory contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
 
 ## Running
 
--- a/examples/phoenix/README.md
+++ b/examples/phoenix/README.md
@@ -1,6 +1,6 @@
 # Phoenix LiveView examples
 
-This directory contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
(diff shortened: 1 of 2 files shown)


examples/phoenix/README.md
```diff
@@ -1,6 +1,6 @@
 # Phoenix LiveView examples
 
-This directory contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
+The [examples/phoenix](https://github.com/elixir-nx/bumblebee/tree/main/examples/phoenix) directory contains contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
```
polvalente, line 3: Since there's even a .md file there, can't we add the examples directory to the release and link to the actual .md file there?

josevalim, line 3:
```suggestion
The [examples/phoenix](https://github.com/elixir-nx/bumblebee/tree/main/examples/phoenix) directory contains minimal, single-file LiveView applications, which showcase how to integrate Bumblebee as part of the application.
```

mix.exs
```diff
@@ -63,7 +63,8 @@ defmodule Bumblebee.MixProject do
         "notebooks/stable_diffusion.livemd",
         "notebooks/llms.livemd",
         "notebooks/llms_rag.livemd",
-        "notebooks/fine_tuning.livemd"
+        "notebooks/fine_tuning.livemd",
+        "examples/phoenix/README.md"
```
josevalim, line 67: I believe new ExDoc versions (or main) support redirects, so we could have a link that points directly to the GitHub repo. Either way is fine by me, your call.

josevalim (approved):

mix.exs
jonatanklosko: I think a redirect would be ideal, but I couldn't find such option.

examples/phoenix/README.md
jonatanklosko: You mean in the archive published to hex? How does this relate to rendering the docs?

polvalente: I meant that if the file is included, I believe it can be referenced as a relative path, but ai might just be wrong

PR #405, opened 2024-11-05
feat: add mpnet model family

snewcomer:
https://arxiv.org/pdf/2004.09297

Huggingface cards
https://huggingface.co/microsoft/mpnet-base
https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-cos-v1

### DIFF
--- a/lib/bumblebee/text/mpnet.ex
+++ b/lib/bumblebee/text/mpnet.ex
@@ -143,6 +143,10 @@ defmodule Bumblebee.Text.MpNet do
     do: [
       :base,
       :for_masked_language_modeling,
+      # :for_sequence_classification, # https://github.com/huggingface/transformers/blob/main/src/transformers/models/mpnet/modeling_mpnet.py
+      # :for_token_classification,
+      # :for_question_answering,
+      # :for_multiple_choice,
     ]
 
   @impl true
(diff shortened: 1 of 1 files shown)


test/bumblebee/text/mpnet_test.exs
```diff
@@ -0,0 +1,51 @@
...
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 32}
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [[-0.2331, 1.7817, 1.1736], [-1.1001, 1.3922, -0.3391], [0.0408, 0.8677, -0.0779]]
+      ])
+    )
```
jonatanklosko, line 28: We want to compare against the reference values from hf/transformers:

```python
from transformers import MPNetModel
import torch

model = MPNetModel.from_pretrained("hf-internal-testing/tiny-random-MPNetModel")

inputs = {
  "input_ids": torch.tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]]),
  "attention_mask": torch.tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])
}

outputs = model(**inputs)

print(outputs.last_hidden_state.shape)
print(outputs.last_hidden_state[:, 1:4, 1:4])

#=> torch.Size([1, 10, 64])
#=> tensor([[[ 0.0033, -0.2547,  0.4954],
#=>          [-1.5348, -1.5433,  0.4846],
#=>          [ 0.7795, -0.3995, -0.9499]]], grad_fn=<SliceBackward0>)
```

I believe there are a few differences between MPNet and BERT, so we need to align the implementation accordingly. In particular, by a quick look some layer names differ, for example `key->k`, `value->v`, `query->q`, so we need to update the layer mapping as well :)

lib/bumblebee/text/mpnet.ex
```diff
@@ -0,0 +1,458 @@
+defmodule Bumblebee.Text.MPNet do
```
jonatanklosko, line 1: Nitpick: I think we want the name to be `MpNet` to align with our naming conventions. Basically, in acronyms we capitalize only the first letter, as in BERT -> `Bert`, `RoBERTa` -> `Roberta`. And we capitalize on each word, such as `ResNet`, `ConvNext`. We do this, because the reference names are often arbitrarily capitalized, and it's not ergonomic for library users to know the exact capitalization.

test/bumblebee/text/mpnet_test.exs
jonatanklosko: Yes! And the Bert implementation is https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/modeling_bert.py, which may be helpful for differences.

PR #407, opened 2024-11-08
Add JinaBert model

joelpaulkoch:
I want to share my work on the JinaBert model.
Not sure if you want to include it at all, since it's not officially part of `transformers`, you must specify `trust_remote_code=True` when running it with `transformers`, and there is still an [open issue](https://github.com/huggingface/transformers/issues/27035).

This PR would enable bumblebee users to run the [jina embeddings v2](https://huggingface.co/jinaai/jina-embeddings-v2-base-en) models.

The implementation of JinaBert is [here](https://huggingface.co/jinaai/jina-bert-implementation/tree/main).

Another issue with this being a custom implementation is that there is another variant that I started to work on: [jinaai/jina-embeddings-v2-base-code](https://huggingface.co/jinaai/jina-embeddings-v2-base-code).

Both, `jinaai/jina-embeddings-v2-base-en` and `jinaai/jina-embeddings-v2-base-code`, specify `JinaBertForMaskedLM` as architecture but point to different implementations.

```
 "_name_or_path": "jinaai/jina-bert-implementation",
  "architectures": [
    "JinaBertForMaskedLM"
  ],
  "auto_map": {
    "AutoConfig": "jinaai/jina-bert-implementation--configuration_bert.JinaBertConfig",
    "AutoModelForMaskedLM": "jinaai/jina-bert-implementation--modeling_bert.JinaBertForMaskedLM",
    "AutoModel": "jinaai/jina-bert-implementation--modeling_bert.JinaBertModel",
    "AutoModelForSequenceClassification": "jinaai/jina-bert-implementation--modeling_bert.JinaBertForSequenceClassification"
  },
```
vs. 
```
  "_name_or_path": "jinaai/jina-bert-v2-qk-post-norm",
  "architectures": [
    "JinaBertForMaskedLM"
  ],
  "auto_map": {
    "AutoConfig": "jinaai/jina-bert-v2-qk-post-norm--configuration_bert.JinaBertConfig",
    "AutoModel": "jinaai/jina-bert-v2-qk-post-norm--modeling_bert.JinaBertModel",
    "AutoModelForMaskedLM": "jinaai/jina-bert-v2-qk-post-norm--modeling_bert.JinaBertForMaskedLM",
    "AutoModelForSequenceClassification": "jinaai/jina-bert-v2-qk-post-norm--modeling_bert.JinaBertForSequenceClassification"
  },
```

Is there a mechanism in `bumblebee` to distinguish these?

There are still some issues in this PR, I will add comments and can work on them over the next days/weeks.

### DIFF
--- a/lib/bumblebee.ex
+++ b/lib/bumblebee.ex
@@ -150,6 +150,8 @@ defmodule Bumblebee do
     "GPTNeoXForCausalLM" => {Bumblebee.Text.GptNeoX, :for_causal_language_modeling},
     "GPTNeoXForSequenceClassification" => {Bumblebee.Text.GptNeoX, :for_sequence_classification},
     "GPTNeoXForTokenClassification" => {Bumblebee.Text.GptNeoX, :for_token_classification},
+    "JinaBertForMaskedLM" => {Bumblebee.Text.JinaBert, :for_masked_language_modeling},
+    "JinaBertModel" => {Bumblebee.Text.JinaBert, :base},
     "LayoutLMForMaskedLM" => {Bumblebee.Multimodal.LayoutLm, :for_masked_language_modeling},
     "LayoutLMForQuestionAnswering" => {Bumblebee.Multimodal.LayoutLm, :for_question_answering},
     "LayoutLMForSequenceClassification" =>
--- a/lib/bumblebee/text/jina_bert.ex
+++ b/lib/bumblebee/text/jina_bert.ex
@@ -71,7 +71,7 @@ defmodule Bumblebee.Text.JinaBert do
     ] ++ Shared.common_options([:use_cross_attention, :num_labels, :id_to_label])
 
   @moduledoc """
-  BERT model family.
+  Jina adaption of BERT model family.
 
   ## Architectures
 
@@ -432,9 +432,7 @@ defmodule Bumblebee.Text.JinaBert do
     decoder? = Keyword.get(opts, :decoder?, false)
 
     embeddings =
-      embedder(inputs["input_ids"], inputs["position_ids"], inputs["token_type_ids"], spec,
-        name: "embedder"
-      )
+      embedder(inputs["input_ids"], inputs["token_type_ids"], spec, name: "embedder")
 
     encoder_outputs =
       encoder(
@@ -462,13 +460,8 @@ defmodule Bumblebee.Text.JinaBert do
     }
   end
 
-  defp embedder(input_ids, position_ids, token_type_ids, spec, opts) do
+  defp embedder(input_ids, token_type_ids, spec, opts) do
     name = opts[:name]
-
-    position_ids =
-      Layers.default position_ids do
-        Layers.default_position_ids(input_ids)
-      end
 
     token_type_ids =
       Layers.default token_type_ids do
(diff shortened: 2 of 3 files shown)


lib/bumblebee/text/jina_bert.ex
```diff
@@ -0,0 +1,565 @@
...
+    cross_attention? = decoder? and spec.use_cross_attention
+
+    alibi_relative_bias_matrix =
+      Axon.nx(hidden_state, fn hidden_state ->
+        {_, seqlen, _} = Nx.shape(hidden_state)
+
+        matrix = alibi_matrix(spec.num_attention_heads, spec.max_positions)
+
+        matrix[[.., .., 0..(seqlen - 1), 0..(seqlen - 1)]]
+      end)
```
joelpaulkoch, line 354: This way, we're recalculating the matrix on each run, right?

In the original implementation, they are storing the matrix and only recalculating when it's too small for the current `seqlen`, otherwise they cut it down to match the dimensions.
I guess we could do the same, maybe using `Axon.StatefulOutput`?

I'm also wondering, shouldn't we know `seqlen` anyways when we're compiling the model?

lib/bumblebee/text/jina_bert.ex
```diff
@@ -0,0 +1,565 @@
...
+        "language_modeling_head.output" => "cls.predictions.decoder",
+        "language_modeling_head.bias" => "cls.predictions",
+        "next_sentence_prediction_head.output" => "cls.seq_relationship",
+        "sequence_classification_head.output" => "classifier",
+        "token_classification_head.output" => "classifier",
+        "multiple_choice_head.output" => "classifier",
+        "question_answering_head.output" => "qa_outputs",
+        "encoder.blocks.{n}.ffn.wo" => "encoder.layer.{n}.mlp.wo",
+        "encoder.blocks.{n}.ffn.layernorm" => "encoder.layer.{n}.mlp.layernorm",
+        "encoder.blocks.{n}.ffn.gated_layers" => "encoder.layer.{n}.mlp.gated_layers"
```
joelpaulkoch, line 561: I took over the param names, so we might want to change them

test/bumblebee/text/jina_bert_test.exs
```diff
@@ -0,0 +1,84 @@
...
+               [-0.1346, 0.1457, 0.5572],
+               [-0.1383, 0.1412, 0.5643],
+               [-0.1125, 0.1354, 0.5599]
+             ])
+           )
+  end
+
+  @tag :skip
+  test ":base" do
+    repo = {:hf, "doesnotexist/tiny-random-JinaBert"}
```
joelpaulkoch, line 37: I've tried to make a `tiny-random-JinaBert` with no success, I might try again..

jonatanklosko:
Hey @joelpaulkoch, thanks for the PR and a great article!

To be honest, I am hesitant to support implementations from the Hub because (a) theoretically they are less stable, because they may be still subject to tweaks; (b) model proliferation is more likely, the `jina-embeddings-v2-base-en` vs `jina-embeddings-v2-base-code` is a good example.

We generally wait until models make it to hf/transformers, though from https://github.com/huggingface/transformers/issues/27035 it's not clear if that's ever going to happen.

At the moment, I would defer the decision and see how the status quo evolves. People can still use the model by installing bumblebee as `{:bumblebee, github: "joelpaulkoch/jina-embeddings-v2"}`.

lib/bumblebee.ex
```diff
@@ -150,6 +150,8 @@ defmodule Bumblebee do
     "GPTNeoXForCausalLM" => {Bumblebee.Text.GptNeoX, :for_causal_language_modeling},
     "GPTNeoXForSequenceClassification" => {Bumblebee.Text.GptNeoX, :for_sequence_classification},
     "GPTNeoXForTokenClassification" => {Bumblebee.Text.GptNeoX, :for_token_classification},
+    "JinaBertForMaskedLM" => {Bumblebee.Text.JinaBert, :for_masked_language_modeling},
```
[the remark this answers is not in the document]
jonatanklosko: There is `JinaBertForMaskedLM` implementation and it has the expected layers. I think the issue is that the model on the hub is actually `JinaBertmodel` and the config is wrong.

So the correct way to workaround this would be specifying architecture when loading:

```elixir
Bumblebee.load_model({:hf, "..."}, architecture: :base)
```

It may be worth opening a PR on the HF repo, changing it to `JinaBertmodel`. Unfortunately, the same is the case for the other checkpoints of this model (small, etc).

lib/bumblebee/text/jina_bert.ex
```diff
@@ -0,0 +1,565 @@
...
+      get_slopes_power_of_2(n_heads)
+    else
+      closest_power_of_2 = 2 ** round(:math.floor(:math.log2(n_heads)))
+
+      get_slopes_power_of_2(closest_power_of_2) ++
+        (get_alibi_head_slopes(2 * closest_power_of_2)
+         |> Enum.take_every(2)
+         |> Enum.take(n_heads - closest_power_of_2))
+    end
+  end
```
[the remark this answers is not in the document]
jonatanklosko: In terms of performance this is actually better than writing a `defn`, assuming `n_heads` is small (which is the case). The reason is that all of this code runs when building the defn expression (defn-compile time), and not as part of the inference. In other words, when compiling the model, we are building this small tensor and it gets embedded as a constant into the computation.

To make the distinction more clear, I would make `alibi_matrix ` a `defnp`, and make `alibi_head_slopes` a `deftransformp` that returns a tensor.

lib/bumblebee/text/jina_bert.ex
jonatanklosko: > I'm also wondering, shouldn't we know seqlen anyways when we're compiling the model?

Yes! PyTorch models generally accept a dynamically sized inputs and adjust the computation accordingly. In our case, we always know the lengths at model compile time.

Consequently, instead of computing the whole matrix and slicing it, ideally we would compute it only for the known length, to avoid unnecessary work and memory consumption.

As for reusing the matrix, in cases like this we always create the tensor at runtime. We prefer the model to be stateless and also let the XLA compiler optimise across operations. There are certain cases where we need some level of statefulness, such as autoregressive text generation, and we do it with an explicit "cache" output/input (though a whole generation request is still stateless).

PR #409, opened 2024-11-22
Update link to Mistral docs

jrgtt:
Hi, I've noticed that the link to the Mistral documentation was returning a 404, so I am adjusting here to the the new one.


jonatanklosko (approved):
Thank you :)

PR #422, opened 2025-10-04
Add SmolLM3

joelpaulkoch:
Hey, this is the SmolLM3 model from huggingface. It's smol, fully open and supports reasoning, so I figured it would be a nice addition to bumblebee.

I didn't implement YaRN extrapolation.


lib/bumblebee/text/smollm3.ex
```diff
@@ -0,0 +1,609 @@
...
+            "decoder.blocks.0.self_attention.query" => "transformer.layers.0.self_attn.q_proj",
+            "decoder.blocks.0.self_attention.value" => "transformer.layers.0.self_attn.v_proj",
+            "decoder.blocks.0.self_attention_norm" => "transformer.layers.0.input_layernorm",
+            "decoder.blocks.0.self_attention.output" => "transformer.layers.0.self_attn.o_proj",
+            "decoder.blocks.0.ffn.output" => "transformer.layers.0.mlp.down_proj",
+            "decoder.blocks.0.ffn.intermediate" => "transformer.layers.0.mlp.up_proj",
+            "decoder.blocks.0.ffn.gate" => "transformer.layers.0.mlp.gate_proj"
+          }
+
+          Map.merge(mapping, question_answering_mapping)
```
jonatanklosko: The use a different prefix for all layers, so we can probably just do this:

```suggestion
          for {key, value} <- mapping, into: %{} do
            {key, String.replace_leading(value, "model.", "transformer.")}
          end
```

test/bumblebee/text/smollm3_test.exs
```diff
@@ -0,0 +1,142 @@
...
+
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 64}
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [[-0.4167, -0.0137, 0.7160], [-0.2624, -1.1185, -0.3098], [-0.0383, -0.8390, -0.0039]]
+      ])
```
jonatanklosko, line 33: Just double-checking, these values come from Python, right?

lib/bumblebee/text/smollm3.ex
```diff
@@ -0,0 +1,609 @@
...
+          * `%{type: :linear, factor: number()}`
+
+          * `%{type: :dynamic, factor: number()}`
+
+          * `%{type: :llama3, factor: number(), low_frequency_factor: number(), high_frequency_factor: number(), original_max_positions: pos_integer()}`
+
+        For more details see https://www.reddit.com/r/LocalLLaMA/comments/14mrgpr/dynamically_scaled_rope_further_increases
+        """
+      ],
+      no_rope_layers: [
```
jonatanklosko: This naming is very confusing, initially I thought it means not-RoPE, but 1 (true) actually enables RoPE. So I guess it rather means No- and Ro-PE.

One alternative configuration I can think of would be `:rotary_embedding_enabled`, with a list of booleans true/false (and if omitted, defaults to true). We can easily convert the representation when loading the config. What do you think?

On a sidenote, we generally use "block" wherever they say "layer" (because it is a group of whole layers).

jonatanklosko (commented):
Hey @joelpaulkoch, this looks great! I dropped a few small comments and it's good to go :)

lib/bumblebee/text/pre_trained_tokenizer.ex
```diff
@@ -211,6 +211,12 @@ defmodule Bumblebee.Text.PreTrainedTokenizer do
         mask: "<mask>"
       }
     },
+    smollm3: %{
+      special_tokens: %{
+        eos: "<|im_end|>",
+        pad: "<|im_end|>"
+      }
+    },
```
jonatanklosko, line 219: We should also add a simple tokenizer test in https://github.com/elixir-nx/bumblebee/blob/main/test/bumblebee/text/pre_trained_tokenizer_test.exs.

test/bumblebee/text/smollm3_test.exs
joelpaulkoch: yes, coming from Python :) although the [repo config](https://huggingface.co/joelkoch/tiny_smollm3_for_sequence_classification/blob/main/config.json) is so tiny, it's not even hitting the no rope layer case.

As a sidenote, I think next time I'll try to set up a simple validation script with `pythonx` so that it can be reused for contributing model implementations.

lib/bumblebee/text/smollm3.ex
joelpaulkoch: I agree the naming is very confusing, took it [directly from huggingface](https://huggingface.co/HuggingFaceTB/SmolLM3-3B/blob/main/config.json#L56) to see what you are going to suggest, sorry. Also, very confusing that they have `no_rope_layers` **and** `no_rope_layer_interval`.

`:rotary_embedding_enabled` sounds good to me :+1:

joelpaulkoch:
The implementation is basically llama + NoPE support (in the transformer block) + architectures that are supported but missing in llama (i.e. `:for_question_answering` and `:for_token_classification`). So, would you prefer to add the [optional NoPE support](https://github.com/elixir-nx/bumblebee/pull/422/commits/2177e11df759788437f6734f6a0713e2c95407f5#diff-e23a07c6c8499c55ef39ea415b8db4697b5484437d83c0e7b90abf83d9a3a1c1) and [architectures](https://github.com/elixir-nx/bumblebee/pull/422/commits/bc5e0e1bfad9359641d9112ddc6e24a2e6303630#diff-e23a07c6c8499c55ef39ea415b8db4697b5484437d83c0e7b90abf83d9a3a1c1R279) in the llama implementation and map smollm3 to llama?

lib/bumblebee.ex
```diff
@@ -188,6 +188,11 @@ defmodule Bumblebee do
     "RobertaForTokenClassification" => {Bumblebee.Text.Roberta, :for_token_classification},
     "RobertaForCausalLM" => {Bumblebee.Text.Roberta, :for_causal_language_modeling},
     "RobertaModel" => {Bumblebee.Text.Roberta, :base},
+    "SmolLM3" => {Bumblebee.Text.SmolLM3, :base},
```
jonatanklosko:
```suggestion
    "SmolLM3Model" => {Bumblebee.Text.SmolLM3, :base},
```

test/bumblebee/text/smollm3_test.exs
jonatanklosko: Perhaps we can use 2 layers and have one use rope and the other use nope?

You can also reduce vocab_size to make the model smaller.

I generated all the models into the [bumblebee-testing org](https://huggingface.co/bumblebee-testing/models?search=smollm3). They are around 200kb.

I used this script (same as llama, just the extra `no_rope_layers`):

```python
from transformers import SmolLM3Config, SmolLM3Model, SmolLM3ForCausalLM, SmolLM3ForQuestionAnswering, SmolLM3ForSequenceClassification, SmolLM3ForTokenClassification

config = SmolLM3Config(
  vocab_size=1024,
  hidden_size=32,
  num_hidden_layers=2,
  num_attention_heads=4,
  intermediate_size=37,
  hidden_act="gelu",
  hidden_dropout_prob=0.1,
  max_position_embeddings=512,
  type_vocab_size=16,
  is_decoder=False,
  initializer_range=0.02,
  pad_token_id=0,
  no_rope_layers=[0, 1]
)

for c in [SmolLM3Model, SmolLM3ForCausalLM, SmolLM3ForQuestionAnswering, SmolLM3ForSequenceClassification, SmolLM3ForTokenClassification]:
  name = c.__name__
  c(config).save_pretrained(f"bumblebee-testing/tiny-random-{name}", repo_id=f"bumblebee-testing/tiny-random-{name}", push_to_hub=True)
```

You can try those and let me know if there are any issues.

test/bumblebee/text/smollm3_test.exs
```diff
@@ -0,0 +1,169 @@
...
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [
+          [0.256240, -0.424804, -0.137127],
+          [-0.806056, -0.141523, 0.364655],
+          [-0.407146, -1.018769, -1.137962]
+        ]
+      ]),
```
jonatanklosko: Nit: we typically only include 4 decimal places, since the further ones are not relevant for the precision we compare with anyway.

```suggestion
      Nx.tensor([
        [
          [0.2562, -0.4248, -0.1371],
          [-0.8060, -0.1415, 0.3646],
          [-0.4071, -1.0187, -1.1379]
        ]
      ]),
```

test/bumblebee/text/smollm3_test.exs
```diff
@@ -0,0 +1,169 @@
...
+      atol: 1.0e-3,
+      rtol: 1.0e-3
```
jonatanklosko: This is unusual, pretty much all other LLMs work with the default atol `10e-4`. If the numbers are slightly off like that, it often indicates a small difference, like a missing layer norm, layer norm in a different order, or something like that.

Do you know if that deviation is only for the test models, or is it similar for any real checkpoint?

joelpaulkoch: I checked and it's the same for `HuggingFaceTB/SmolLM3-3B`, not sure what is missing.
A missing layer norm would show up in the debug logs, right?
I'll investigate further when I find the time.

jonatanklosko: > A missing layer norm would show up in the debug logs, right?

A norm layer that bumblebee has and cannot be found in the checkpoint would be logged. But if there is a norm layer in the checkpoint that bumblebee doesn't use, then it is not logged, unless we pass `log_params_diff: true` to `Bumblebee.load_model` (sometimes unused layers are expected, for example if you load a checkpoint as a base model, ignoring the head layers). But I just tried with `log_params_diff: true` and it doesn't log anything, so in terms of layers it looks like they match.

jonatanklosko: @joelpaulkoch I had a look and found the issue, see bf4d8d8ea1670b22d2645b2a00cb3c1a0e71f77d. I first checked the config loaded from HF and noticed double nesting in `rotary_embedding_enabled: %{rotary_embedding_enabled: ...}`. And the main issue was reading `:rotary_embedding_enabled` from `opts`, rather than `spec`, which means it was never actually applied:

```diff
-      case opts[:rotary_embedding_enabled] do
+      case spec.rotary_embedding_enabled do
```

Btw. Looking at this, I realised that that rotary embedding layer doesn't have any params, so we don't need the layer mapping:

```elixir
                {"decoder.blocks.#{index}.self_attention.rotary_embedding",
                 "model.layers.#{index}.self_attn.rotary_emb"}
```

I believe it was added as part of llama unnecessarily, and then by copying we kept that across many models, but it doesn't actually do anything. I will remove this mapping everywhere in a separate commit.

joelpaulkoch: oh great, thank you @jonatanklosko !!

PR #425, opened 2025-10-21
Add state to logits processing

xhr15:
Logits processing is a powerful tool, particularly for using smaller language models for tasks such as named entity recognition. @seanmor5 started work in this area with https://github.com/elixir-nx/bumblebee/pull/354. 

Whatever the approach, it will require some kind of state. 

This pull request is a proposal to allow logits processors to be stateful. 

This would enable the use of deterministic finite automata (DFAs) or pushdown automata (PDAs) for processing constrained grammars in logits processing. https://github.com/bitcrowd/bumblebee/pull/6 shows how this would be used. We will follow up on this PR if this approach is favoured.

### DIFF
--- a/lib/bumblebee/logits_processor.ex
+++ b/lib/bumblebee/logits_processor.ex
@@ -16,10 +16,10 @@ defmodule Bumblebee.LogitsProcessor do
   @type state :: Nx.Container.t()
 
   @type process_context :: %{
-    sequence: Nx.Tensor.t(),
-    length: Nx.Tensor.t(),
-    input_length: Nx.Tensor.t()
-  }
+          sequence: Nx.Tensor.t(),
+          length: Nx.Tensor.t(),
+          input_length: Nx.Tensor.t()
+        }
 
   @type init_context :: %{}
 


test/bumblebee/text/generation/logits_processing_test.exs
```diff
@@ -5,6 +5,53 @@ defmodule Bumblebee.Text.Generation.LogitsProcessingTest do
 
   alias Bumblebee.Text.Generation.LogitsProcessing
 
+  describe "stateful logits processors" do
+    defmodule StatefulLogitsProcessing do
```
jonatanklosko: I would skip that test. Here we defined `stateful_processor` for test purposes and then test that function in isolation, so we don't actually test any of the Bumblebee code.

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,149 @@ defmodule Bumblebee.Text.GenerationTest do
...
+
+    # first token_id still 80 as we suppress token_id 79
+    assert_equal(token_ids[[0,0]], 80)
+    # in the next step we increment from 79 to 80 and suppress token_id 80, the
+    #result is 176 as that is the next likelihood in the logits.
+
+    assert_equal(token_ids[[0,1]], 176)
+  end
+
+  test "with stateful logits processor with batch size of 2" do
```
jonatanklosko: I would merge this with the test above:

```elixir
test "with stateful logits processor" do
  ...

  # With batch size 1

  ...

  # With batch size 2

  ...
end
```

This way we load the model only once.

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,149 @@ defmodule Bumblebee.Text.GenerationTest do
...
+
+    # We demonstrate the use of the state with the following example of a
+    # stateful processor (see below). On the first iteration, it suppresses the
+    # given initial ID, then increments the token ID to be suppressed on the
+    # following iterations. The ID of the token to be suppressed is passed on
+    # between iterations using the logits_processor_state.
+    #
+    # So invoked with the initial ID of 79, it suppresses 79, 80, 81, ... in
+    # the subsequent iterations, demonstrating the use of the state in a
+    # logits processor.
```
jonatanklosko: A simpler processor to test it would be to force specific tokens. You can keep a counter in the state and then force subsequent tokens 1, 2, 3. This should make the assertions a bit more direct.

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,149 @@ defmodule Bumblebee.Text.GenerationTest do
...
+        context.logits_processor_state[:next_suppressed_token_id] || initial_suppressed_token_id
+
+      logits = suppress_id(logits, suppressed_id)
+
+      next_suppressed_token_id = Nx.add(suppressed_id, 1)
+
+      context =
+        put_in(
+          context,
+          [:logits_processor_state, :next_suppressed_token_id],
```
jonatanklosko: With the current API, the state is always initialized to `%{}` and then first invocation of the processor adds a key, here `%{next_suppressed_token_id: %Nx.Tensor{...}}`.

This can be problematic in defn `while` loop, which requires the accumulation sate to always have the same shape. In other words, the initial state should already include `:next_suppressed_token_id` with the default tensor. It is possible that this didn't come up during your tests, because depending on the model/input, we do the first generation step outside of the `while` loop, and the first call would initialize the state. However, if we are going to support stateful, I would rather do it in a more robust way.

Given the above, a stateless logits processor would involve two steps (functions):

1. Building an initial state.
2. Performing logits processing, which receives logits and state, and returns update logits and state.

This way we can call (1) when initializing the generation context, and for the actual processing we call (2).

The behaviour can be similar to `Bumblebee.Scheduler`. Something like this:

```elixir
defmodule Bumblebee.LogitsProcessor do
  @moduledoc """
  An interface for configuring and using logits processors.

  Logits processors are used during autoregressive generation to modify
  predicted scores at each generation step. This allows for applying
  certain rules to the model output to control which tokens are picked
  at each generation step, and which are not.

  Every module implementing this behaviour is expected to also define
  a configuration struct.
  """

  @type t :: Bumblebee.Configurable.t()

  @type state :: Nx.Container.t()

  @doc """
  Initializes state for a new logits processor.

  Returns `state`, which is an opaque `Nx.Container`, and it is then
  passed to and returned from `process/2`.

  Oftentimes logits processors are stateless, in which case this
  function can return an empty continer, such as `{}`.
  """
  @callback init(t(), context) :: state()
            when context: %{
                   prng_key: Nx.Tensor.t()
                 }

  @doc """
  Processes logits, applying specific rules.
  """
  @callback process(
              t(),
              state(),
              logits :: Nx.Tensor.t(),
              context :: context
            ) :: {state :: map(), logits :: Nx.Tensor.t()}
            when context: %{
                   sequence: Nx.Tensor.t(),
                   length: Nx.Tensor.t(),
                   input_length: Nx.Tensor.t()
                 }
end
```

Technically, the `:logits_processors` options is public API, but we can make it backward-compatible. For example, we can define `%Bumblebee.Text.Generation.StatelessLogitsProcessor{fun: fun}`, where the state is always empty and process just invokes the fun. I would even use that for the built-in processors, so that we don't need to define a bunch of new modules.

jonatanklosko (commented):
Hey @xhr15 and @joelpaulkoch, thanks for the PR!

I dropped a few comments, but the main one is about the API. I know it's a bit more involved, but probably worth it. Let me know what you think, and if you have any concerns!

test/bumblebee/text/generation_test.exs
xhr15: @jonatanklosko Thank you very much for your comments! I think esp. the two step call makes sense. We'll move in that direction :)

xhr15: @jonatanklosko 
as an afterthought: 

What is the use case for context here:

```
@callback init(t(), context) :: state()
            when context: %{
                   prng_key: Nx.Tensor.t()
                 }
```

Later in the loop, context holds:

```
context = %{
      sequences: sequences,
      input_length: length,
      length: length,
    }
```

I am wondering how those would influence the initialisation of the logits processors?

Or are you planning of using additional keys? E.g. from the state as returned by init squence:

```
%{
      sequences: sequences,
      input_length: length,
      length: length,
      finished_length: finished_length,
      ignored: Nx.broadcast(0, {batch_size})
    }
```

If that was the case, we should probably rename the parameter to `state` or `initial_state`.

Wdyt?

jonatanklosko: > What is the use case for context here:

I picked "context" in both functions as a generic name for state/metadata that may be relevant to the logits processor. You can see that in my snippet the context type is different for `init` and `process`. Technically all of the context fields could be separate arguments, but keeping it as a map makes the signature more manageable, and more importantly allows us to add more fields in the future without breaking compatibility.

Does that make sense?

lib/bumblebee/text/generation.ex
```diff
@@ -386,18 +388,39 @@ defmodule Bumblebee.Text.Generation do
...
+        if is_function(processor, 2) do
+          %Bumblebee.Text.Generation.StatelessLogitsProcessor{fun: processor}
+        else
+          processor
+        end
+      end)
+
+    init_fun = fn context ->
+      Enum.reduce(processors, %{}, fn processor, state_acc ->
+        state = Bumblebee.logits_processor_init(processor, context)
```
jonatanklosko: Currently each processor needs to be careful to namespace its state and avoid conflicts.  Ideally the processor would not need to care about it, so in your example:

```diff
def init(logits_processor, _context) do
  initial_enforced_token_ids =
    Enum.map(logits_processor.initial_enforced_token_ids, &List.wrap(&1))

  initial_enforced_batch_token_id =
    Nx.tensor(initial_enforced_token_ids)

-  %{
-    sfp_state: %{
-      next_enforced_token_id: initial_enforced_batch_token_id
-    }
-  }
+  %{
+    next_enforced_token_id: initial_enforced_batch_token_id
+  }
end
```

To do this, instead of having a single map and `Map.merge` into it, we can instead have a list of processor states. We init them separately, and we zip processors with their states for updates. Something like this:

```elixir
init_fun = fn context ->
  processors
  |> Enum.map(fn processor ->
    Bumblebee.logits_processor_init(processor, context)
  end)
  |> List.to_tuple()
end

process_fun = fn logits, context, processor_states ->
  {processor_states, logits} =
    processors
    |> Enum.zip(Tuple.to_list(processor_states))
    |> Enum.map_reduce(logits, fn {processor, processor_state}, logits ->
      Bumblebee.logits_processor_process(processor, processor_state, logits, context)
    end)
    
  {List.to_tuple(processor_states), logits}
end
```

Note that we want to keep the states as a tuple instead of list, so that it is a valid Nx container and can be passed to `while` and around `defn` calls.

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,165 @@ defmodule Bumblebee.Text.GenerationTest do
...
+
+    # this is the same example as above, but with a batch size of 2.
+
+    generation_config = Bumblebee.configure(generation_config, max_new_tokens: 3)
+
+    generate =
+      Bumblebee.Text.Generation.build_generate(model, spec, generation_config,
+        logits_processors: [
+          Bumblebee.configure(Bumblebee.Text.GenerationTest.StatefulLogitsProcessing,
+            initial_enforced_token_ids: [78, 20]
```
jonatanklosko: This is an arbitrary example, but I don't think we would ever have different value for each entry in the batch, because the batch entries should generally be interchangable. So it should rather be `initial_enforced_token_id: 78`, and then you check that it is enforced in the same way for both batch entries.

jonatanklosko:
> We might not want to vectorize all the state of the logits processors e.g. when we want to read from a shared state tensor while processing the vectorized logits we would otherwise have to duplicate the shared state tensor across the vectorized axis, right? We can instead vectorize only the state that needs vectorization inside the logits processor.
> 
> That's basically the reason for [2ba5e0a](https://github.com/elixir-nx/bumblebee/commit/2ba5e0adb32eaeda42a957a38048363acc21ea57), I'm not entirely sure if this is alright or if it has negative implications for `defn`.

Correct, Bumblebee should not call vectorize on the logits processor state. Ideally we want vectorization to happen automatically.

For example, schedulers have a similar init, here's one of them:

https://github.com/elixir-nx/bumblebee/blob/bc1b4525620e9afe4b2542324b3520762e37fdc5/lib/bumblebee/diffusion/pndm_scheduler.ex#L97-L108

`alpha_bars` is generated as a flat tensor and it is *shared state* (not duplicated across batch). On the other hand, the caller (Bumblebee) can pass `sample_template` with vectorized axis and then `empty = Nx.fill(sample_template, 0)` would be vectorized state. What's nice is that the scheduler is not aware about the vectorization, and a non-vectoriezd input works just fine.

For this to work automatically though, we need something to derive state of off (like `sample_template`), so that it gets automatically vectorized. I'm not yet sure how it would look for the processor, I need to think more about this.

jonatanklosko:
Sorry for the late reply, I was off last week :)

lib/bumblebee/text/generation.ex
xhr15: Thank you @jonatanklosko, we were wondering if having this kind of interface to reach into other processors would be a good or bad thing :) 

We'll take it out!

lib/bumblebee/logits_processor.ex
```diff
@@ -0,0 +1,38 @@
...
+  @callback init(t(), any()) :: state()
+
+  @doc """
+  Processes logits, applying specific rules.
+  """
+  @callback process(
+              t(),
+              state(),
+              logits :: Nx.Tensor.t(),
+              context :: term()
```
jonatanklosko: Let's use a more precise type. We could defined it inline, but since we reference it in `Bumblebee.logits_processor_process`, we should rather defined a type:

```elixir
@type process_context: %{
        sequence: Nx.Tensor.t(),
        length: Nx.Tensor.t(),
        input_length: Nx.Tensor.t()
      }
```

Same for init, for now we can define it as `@type init_context :: %{}`, and we add fields if needed.

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,160 @@ defmodule Bumblebee.Text.GenerationTest do
...
+    def config(logits_processor, opts) do
+      Bumblebee.Shared.put_config_attrs(logits_processor, opts)
+    end
+
+    @impl Bumblebee.LogitsProcessor
+    def init(logits_processor, context) do
+      initial_enforced_token_id = Nx.tensor([logits_processor.initial_enforced_token_id])
+
+      [initial_enforced_batch_token_id, _sequence] =
+        Nx.broadcast_vectors([initial_enforced_token_id, context.sequence])
```
jonatanklosko: Technically this state doesn't need to be vectorized, since the value is always the same for every batch item?

test/bumblebee/text/generation_test.exs
```diff
@@ -106,4 +106,160 @@ defmodule Bumblebee.Text.GenerationTest do
...
+
+    # in the next step we increment from 79 to 80 and enforce token_id 80
+    assert_equal(token_ids[[0, 2]], 80)
+
+    # second entry in batch
+    # first token_id is 20 as we enforce token_id 20 on the first iteration
+    assert_equal(token_ids[[1, 0]], 78)
+
+    # in the next step we increment from 20 to 21 and enforce token_id 21
+    assert_equal(token_ids[[1, 1]], 79)
```
jonatanklosko, line 196: I would remove most comments in this test, since they make it seem more complicated than it actually is :D For example here:

```suggestion
    # first entry in batch
    assert_equal(token_ids[[0, 0]], 78)
    assert_equal(token_ids[[0, 1]], 79)
    assert_equal(token_ids[[0, 2]], 80)

    # second entry in batch
    assert_equal(token_ids[[1, 0]], 78)
    assert_equal(token_ids[[1, 1]], 79)
    assert_equal(token_ids[[1, 2]], 80)
```

test/bumblebee/text/generation_test.exs
xhr15: 👍 good catch!

notebooks/debug_print.livemd
```diff
@@ -0,0 +1,203 @@
...
+generate_fun =
+  Bumblebee.Text.Generation.build_generate(model, spec, generation_config,
+    logits_processors: [ &PrintLogitsProcessor.debug_processor(&1, &2, [debug_limit: 2])]
+  )
+```
+
+## Setting up the serving
+
+This is taken from `lib/bumblebee/text/text_generation.ex`. It's basically the `Bumblebee.Text.generation` function that you usually use to create text generation servings (with some minor modifications to simplify it).
+We must use the lower level API here to be able to include `PrintLogitsProcessor` in `generate_fun`.
```
jonatanklosko: Oh, I think it's fine to add the logits processors argument to the serving, especially now that we have a whole API around it.

notebooks/debug_print.livemd
jonatanklosko, line 1: Perhaps these two notebooks should be a single notebook called "logits processing", or "constraining generation output".

test/bumblebee/text/generation/logits_processing_test.exs
```diff
@@ -382,7 +382,8 @@ defmodule Bumblebee.Text.Generation.LogitsProcessingTest do
     %{
       sequence: Nx.tensor(sequence),
       length: Enum.count(sequence, &(&1 != 0)),
-      input_length: 1
+      input_length: 1,
+      logits_processor_state: %{}
```
jonatanklosko: This is not part of the context anymore.

```suggestion
      input_length: 1
```

xhr15: Thank you! :)

notebooks/debug_print.livemd
xhr15: Sure. I'll take the livebooks out of this PR, then create a new one once this is merged

notebooks/debug_print.livemd
xhr15: Yes, was thinking about that as well. Will do!

jonatanklosko (approved):
Very nice!

lib/bumblebee/logits_processor.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @type t :: Bumblebee.Configurable.t()
+
+  @type state :: Nx.Container.t()
+
+  @type process_context :: %{
+    sequence: Nx.Tensor.t(),
+    length: Nx.Tensor.t(),
+    input_length: Nx.Tensor.t()
+  }
```
jonatanklosko: CI fails formatting:

```suggestion
  @type process_context :: %{
          sequence: Nx.Tensor.t(),
          length: Nx.Tensor.t(),
          input_length: Nx.Tensor.t()
        }
```

Issue #428, opened 2025-12-17
Bumblebee.Vision.DinoV2 crashes when batch_size > 1


zacharydenton:
I'm trying to use Bumblebee.Vision.DinoV2 in an Nx.Serving. It crashes when I use Nx.Serving.batched_run():

```
# DinoV2Serving is a Bumblebee.Vision.image_embedding() serving
iex(1)> image = Nx.broadcast(0.5, {64, 64, 3})
iex(2)> for i <- 1..10, do: Task.async fn -> Nx.Serving.batched_run(DinoV2Serving, image) end
    ** (EXIT) an exception was raised:
        ** (Axon.CompileError) exception found when compiling layer anonymous fn/3 in Bumblebee.Vision.DinoV2.interpolate_position_embeddings/3 named interpolate_position_embeddings_0:

    ** (ArgumentError) cannot reshape, current shape {1, 225, 32} is not compatible with new shape {10, 15, 15, 32}
        (nx 0.10.0) lib/nx/shape.ex:184: Nx.Shape.reshape/2
        (nx 0.10.0) lib/nx.ex:2832: Nx.reshape/3
        (bumblebee 0.6.3) lib/bumblebee/vision/dino_v2.ex:324: anonymous fn/4 in Bumblebee.Vision.DinoV2.interpolate_position_embeddings/3


(pass debug: true to build/compile to see where the layer was defined)


Compiling of the model was initiated at:

            (bumblebee 0.6.3) lib/bumblebee/vision/image_embedding.ex:37: anonymous fn/6 in Bumblebee.Vision.ImageEmbedding.image_embedding/3
            (nx 0.10.0) lib/nx/defn/compiler.ex:163: Nx.Defn.Compiler.runtime_fun/3
            (nx 0.10.0) lib/nx/defn/evaluator.ex:105: Nx.Defn.Evaluator.precompile/3
            (nx 0.10.0) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.__compile__/4
            (nx 0.10.0) lib/nx/defn/evaluator.ex:48: Nx.Defn.Evaluator.__jit__/5
            (nx 0.10.0) lib/nx/defn/compiler.ex:134: Nx.Defn.Compiler.__jit__/4
```

PR #429, opened 2025-12-17
Fix DinoV2 crash when batch_size > 1

zacharydenton:
This fixes an error in interpolate_position_embeddings when batch_size is > 1 (#428).

### DIFF
--- a/lib/bumblebee/vision/dino_v2.ex
+++ b/lib/bumblebee/vision/dino_v2.ex
@@ -322,7 +322,10 @@ defmodule Bumblebee.Vision.DinoV2 do
 
         interpolated_position_embeddings =
           input_position_embeddings
-          |> Nx.reshape({position_embeddings_batch_size, original_positions, original_positions, spec.hidden_size})
+          |> Nx.reshape(
+            {position_embeddings_batch_size, original_positions, original_positions,
+             spec.hidden_size}
+          )
           |> Axon.Layers.resize(
             size: {resized_height, resized_width},
             method: :bicubic,


lib/bumblebee/vision/dino_v2.ex
```diff
@@ -321,13 +321,13 @@ defmodule Bumblebee.Vision.DinoV2 do
 
         interpolated_position_embeddings =
           input_position_embeddings
-          |> Nx.reshape({batch_size, original_positions, original_positions, spec.hidden_size})
+          |> Nx.reshape({1, original_positions, original_positions, spec.hidden_size})
```
jonatanklosko: @zacharydenton great find! I would suggest a small improvement:

```elixir
position_embeddings_batch_size = Nx.axis_size(position_embeddings, 0)
```

and then use that in both places.

In practice it's always `1`, but hardcoding the value implies an extra assumption that the reader may need to understand, which means going back to the caller side and figuring out the position embeddings shape :)

zacharydenton: Done.

lib/bumblebee/vision/dino_v2.ex
```diff
@@ -312,22 +312,23 @@ defmodule Bumblebee.Vision.DinoV2 do
...
         resized_width = div(width, spec.patch_size)
+        position_embeddings_batch_size = Nx.axis_size(position_embeddings, 0)
 
         class_position_embedding = position_embeddings[[.., 0..0//1, ..]]
         input_position_embeddings = position_embeddings[[.., 1..-1//1, ..]]
 
         interpolated_position_embeddings =
           input_position_embeddings
-          |> Nx.reshape({batch_size, original_positions, original_positions, spec.hidden_size})
+          |> Nx.reshape({position_embeddings_batch_size, original_positions, original_positions, spec.hidden_size})
```
jonatanklosko:
```suggestion
          |> Nx.reshape(
            {position_embeddings_batch_size, original_positions, original_positions,
             spec.hidden_size}
          )
```

PR #430, opened 2025-12-18
Add InceptionResnetV1 model.

zacharydenton:
This is a port of https://github.com/timesler/facenet-pytorch/blob/master/models/inception_resnet_v1.py.

I'm using this to generate facenet512 embeddings for face recognition:

```elixir
{:ok, facenet_model_info} =
  Bumblebee.load_model({:hf, "zacharydenton/facenet-vggface2"}, module: Bumblebee.Vision.InceptionResnetV1))

{:ok, facenet_featurizer} =
  Bumblebee.load_featurizer({:hf, "zacharydenton/facenet-vggface2"})

facenet_serving =
  Bumblebee.Vision.image_embedding(facenet_model_info, facenet_featurizer,
    embedding_processor: :l2_norm,
  )

# use face detection (e.g. yunet from evision) to get x/y/w/h
face_img = extract_face!(img, x, y, w, h)
{:ok, facenet512} = Nx.Serving.batched_run(FacenetServing, face_img)
```

I've tested this implementation and compared to the embeddings generated by the facenet-pytorch library. The embeddings match the pytorch implementation with 0.9995 average cosine similarity (with latest Axon - see https://github.com/elixir-nx/axon/pull/622 for details).

n.b. I want to add unit tests but I didn't find any suitable weights in https://huggingface.co/hf-internal-testing - open to advice!

### DIFF
--- a/lib/bumblebee/vision/inception_resnet_v1.ex
+++ b/lib/bumblebee/vision/inception_resnet_v1.ex
@@ -75,7 +75,8 @@ defmodule Bumblebee.Vision.InceptionResnetV1 do
   @impl true
   def input_template(spec) do
     %{
-      "pixel_values" => Nx.template({1, spec.image_size, spec.image_size, spec.num_channels}, :f32)
+      "pixel_values" =>
+        Nx.template({1, spec.image_size, spec.image_size, spec.num_channels}, :f32)
     }
   end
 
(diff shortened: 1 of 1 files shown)


lib/bumblebee/vision/inception_resnet_v1.ex
```diff
@@ -0,0 +1,369 @@
...
+
+    pooled_state =
+      input
+      |> stem(spec, name: join(name, "stem"))
+      |> inception_resnet_blocks(spec, name: join(name, "blocks"))
+      |> head(spec, name: join(name, "head"))
+
+    %{
+      pooled_state: pooled_state,
+      hidden_states: Axon.container({input, pooled_state})
```
zacharydenton, line 118: I'm not sure what to do with `hidden_state` and `hidden_states`

jonatanklosko: `hidden_states` is a tuple (effectively a list) of intermediate states between various blocks (see `Bumblebee.Vision.ResNet` for example). By default we actually prune it and only return if the user configures `output_hidden_states: true`. We conventionally do it for most models to mirror hf/transformers, but is not a requirement, and it should be fine to not return it.

PR #431, opened 2025-12-19
Add Mistral3 multimodal support with Pixtral vision encoder

dai-yamashita:
## Summary

This PR adds support for Mistral3 multimodal models (Ministral series):

- **Mistral text model extensions**: `attention_head_size`, `use_interleaved_attention`, `tie_word_embeddings`
- **Pixtral vision encoder**: `Bumblebee.Vision.Pixtral`
- **Mistral3 text decoder**: `Bumblebee.Text.Mistral3`
- **Multimodal model**: `Bumblebee.Multimodal.Mistral3`

### Supported architectures
- `PixtralVisionModel`
- `Mistral3Model`, `Mistral3ForCausalLM`, `Mistral3ForSequenceClassification`
- `Mistral3ForConditionalGeneration` (multimodal)

### Key features

1. **Function-based attention_window_size** in `transformer.ex` for per-layer attention configuration
2. **Interleaved attention pattern**: even layers use global attention, odd layers use sliding window
3. **Multimodal projector**: patch merger + linear layers to project vision features to text embedding space

## Test plan
- [x] All existing tests pass (264 tests, 0 failures)
- [x] New tests for Pixtral, Mistral3, and Multimodal added


lib/bumblebee.ex
```diff
@@ -170,6 +170,13 @@ defmodule Bumblebee do
     "MistralModel" => {Bumblebee.Text.Mistral, :base},
     "MistralForCausalLM" => {Bumblebee.Text.Mistral, :for_causal_language_modeling},
     "MistralForSequenceClassification" => {Bumblebee.Text.Mistral, :for_sequence_classification},
+    "Mistral3Model" => {Bumblebee.Text.Mistral3, :base},
+    "Mistral3ForCausalLM" => {Bumblebee.Text.Mistral3, :for_causal_language_modeling},
+    "Mistral3ForSequenceClassification" =>
+      {Bumblebee.Text.Mistral3, :for_sequence_classification},
```
jonatanklosko, line 176: As far as I can tell `Mistral3ForCausalLM` and `Mistral3ForSequenceClassification` don't exist in hf/transformers. Looking at [modeling_mistral3.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral3/modeling_mistral3.py) there's only `Mistral3ForConditionalGeneration` and the corresponding `Mistral3Model`, which should be a single module in Bumblebee.

test/bumblebee/multimodal/mistral3_test.exs
```diff
@@ -0,0 +1,98 @@
...
+    assert Map.has_key?(outputs, :logits)
+    assert Nx.shape(outputs.logits) == {1, 5, 1024}
+
+    # Expected values from Bumblebee inference with model generated by:
+    # test/fixtures/scripts/generate_expected_values.py (torch.manual_seed(42))
+    assert_all_close(
+      outputs.logits[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [
+          [3.5014779567718506, -3.962040662765503, -4.744167327880859],
```
jonatanklosko, line 34: We should not assert against values from Bumblebee, because it doesn't verify the current implementation is correct. We want those reference values to be generated using Python transformers, this way we know the implementation behaves the same. See #422 for a example complete PR with tests.

dai-yamashita: @jonatanklosko 

Thank you very much for your review.
We truly appreciate your thoughtful advice on various pending issues and how to proceed.

We will carefully address the points you raised and resubmit the request accordingly.

PR #434, opened 2025-12-27
feat: Add MPNet model family

georgeguimaraes:
Adds the MPNet model family to Bumblebee.

MPNet (Masked and Permuted Pre-training) is an encoder model that combines masked language modeling with permuted language modeling.

## Supported architectures

- `:base`
- `:for_masked_language_modeling`
- `:for_sequence_classification`
- `:for_token_classification`
- `:for_question_answering`
- `:for_multiple_choice`

## Implementation notes

- Uses relative position bias shared across all encoder layers
- Does not use `token_type_ids` (unlike BERT/RoBERTa)
- Position IDs start at `pad_token_id + 1` (matching Python transformers)

## Testing

- All 6 architecture tests pass with tiny-random models
- Tokenizer test added with expected values from Python
- Verified output matches Python transformers exactly with `microsoft/mpnet-base`
- Tested sequence classification with real model (`moshew/mpnet-base-sst2-distilled`)

## References

- [MPNet paper](https://arxiv.org/abs/2004.09297)
- [HuggingFace MPNet](https://huggingface.co/docs/transformers/model_doc/mpnet)


test/bumblebee/text/mpnet_test.exs
```diff
@@ -0,0 +1,149 @@
...
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 64}
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [[-0.7203, -1.2364, -0.1180], [-1.1624, -1.0586, 0.1338], [0.7040, 1.3575, 0.4602]]
+      ])
+    )
```
jonatanklosko, line 28: In all the tests we should assert against values from Python transformers, see https://github.com/elixir-nx/bumblebee/pull/405#discussion_r1830459618. If it doesn't match, which seems to be the case here, it means there is still some implementation difference.

georgeguimaraes: got it! will do!

lib/bumblebee/text/mpnet.ex
```diff
@@ -0,0 +1,485 @@
+defmodule Bumblebee.Text.Mpnet do
```
jonatanklosko: To match our conventional naming, let's call this `MpNet`. Basically for acronym like BERT we only capitalize the first letter (-> Bert), but if part of the name is another word, we capitalize it also, here `net`. For example, other models we have are `ControlNet`, `UNet`.

georgeguimaraes: done

PR #435, opened 2025-12-28
Add ModernBERT family

georgeguimaraes:
This adds support for ModernBERT, a recent encoder model that improves on BERT with a few architectural changes:

- Rotary position embeddings (RoPE) instead of absolute position embeddings
- Alternating local and global attention layers for efficiency on longer sequences
- Gated linear units (GeGLU) in the feed-forward blocks
- Pre-normalization (norm before attention/FFN rather than after)
- No bias in layer normalization

## ModernBert (Encoder)

Supported architectures:

- `:base`
- `:for_masked_language_modeling`
- `:for_sequence_classification`
- `:for_token_classification`

The MLM head uses tied embeddings (shares weights with the input token embeddings).

## ModernBertDecoder (Decoder)

A decoder-only variant trained for causal language modeling (text generation).

Supported architectures:

- `:base`
- `:for_causal_language_modeling`

Reference: https://arxiv.org/abs/2412.13663

### DIFF
--- a/lib/bumblebee/text/modernbert.ex
+++ b/lib/bumblebee/text/modernbert.ex
@@ -313,18 +313,6 @@ defmodule Bumblebee.Text.ModernBert do
         Layers.default_position_ids(hidden_state)
       end
 
-    ffn_fun =
-      &gated_ffn(&1, spec.intermediate_size, spec.hidden_size,
-        activation: spec.activation,
-        name: &2
-      )
-
-    state = %{
-      hidden_state: hidden_state,
-      hidden_states: Axon.container({hidden_state}),
-      attentions: Axon.container({})
-    }
-
     layer_types = spec.layer_types || generate_layer_types(spec.num_blocks)
 
     attention_window_size = fn idx ->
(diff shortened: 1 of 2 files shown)


test/bumblebee/text/modernbert_test.exs
```diff
@@ -0,0 +1,57 @@
...
+
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 32}
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [[-0.4497, -2.436, 0.0269], [0.8374, -1.6001, -0.0694], [0.8867, 0.7041, 0.0353]]
+      ])
```
jonatanklosko, line 27: Same comment as in https://github.com/elixir-nx/bumblebee/pull/434#discussion_r2651088062.

The values I get from Python:

```
tensor([[[ 1.2332, -0.7295,  0.1871],
         [ 0.5687, -0.0640,  0.0617],
         [ 0.3401, -3.6260,  0.0752]]], grad_fn=<SliceBackward0>)
```

test/bumblebee/text/modernbert_test.exs
```diff
@@ -0,0 +1,57 @@
+defmodule Bumblebee.Text.ModernBertTest do
+  use ExUnit.Case, async: true
+
+  import Bumblebee.TestHelpers
+
+  @moduletag model_test_tags()
+
+  # Note: sequence_classification and token_classification tests are skipped
+  # because the tiny-random test models have incompatible head structures.
+  # The architectures work correctly with production models.
```
jonatanklosko: What do you mean by "incompatible head structures"? It should handle tiny models as any other pretrained checkpoint.

test/bumblebee/text/modernbert_test.exs
georgeguimaraes: on it!

lib/bumblebee/text/modernbert.ex
```diff
@@ -0,0 +1,585 @@
...
+        doc: "apply global attention every N layers (1 means every layer is global)"
+      ],
+      local_rope_theta: [
+        default: 10_000.0,
+        doc: "the base for computing rotary embedding frequency for local attention layers"
+      ],
+      global_rope_theta: [
+        default: 160_000.0,
+        doc: "the base for computing rotary embedding frequency for global attention layers"
+      ]
```
jonatanklosko: We should unify the naming, here's what we have in Gemma 3:

https://github.com/elixir-nx/bumblebee/blob/16256442e07d9d85cc77dcce08c115b00b759465/lib/bumblebee/text/gemma3_text.ex#L56-L63

lib/bumblebee/text/modernbert.ex
```diff
@@ -0,0 +1,585 @@
...
+      initializer_scale: [
+        default: 0.02,
+        doc:
+          "the standard deviation of the normal initializer used for initializing kernel parameters"
+      ],
+      local_attention_window: [
+        default: 128,
+        doc: "the window size for local attention layers"
+      ],
+      global_attention_every_n_layers: [
```
jonatanklosko: This is deprecated in transformers v5 RC, they use a list of layer types, see [here](https://github.com/huggingface/transformers/blob/e8c51d1848187b9e58d00bf7d638811686ab2a4b/src/transformers/models/modernbert/configuration_modernbert.py#L209-L216). We need to handle both on loading for backward compatibility with model checkpoints.

You can search for `layer_types` in the [gemma3_test.ex](https://github.com/elixir-nx/bumblebee/blob/main/lib/bumblebee/text/gemma3_text.ex) implementation, we should do the same here.

lib/bumblebee/text/modernbert.ex
```diff
@@ -0,0 +1,585 @@
...
+  defp layer_norm(x, opts) do
+    name = opts[:name]
+    epsilon = opts[:epsilon] || 1.0e-5
+
+    Axon.layer(
+      fn x, gamma, _opts ->
+        mean = Nx.mean(x, axes: [-1], keep_axes: true)
+        variance = Nx.variance(x, axes: [-1], keep_axes: true)
+        normalized = Nx.divide(Nx.subtract(x, mean), Nx.sqrt(Nx.add(variance, epsilon)))
+        Nx.multiply(normalized, gamma)
```
jonatanklosko: For the implementation we can call [Axon.Layers.layer_norm](https://hexdocs.pm/axon/Axon.Layers.html#layer_norm/4) (this is the Nx definition behind the Axon layer). We just pass 0 as beta (bias).

jonatanklosko (approved):
I refactored it to use `Layers.Transformer.blocks`, instead of doing the loop ourselves.

FTR I considered adding `index` as a parameter to the `:block_type` function, so that we could have the conditional there. It would be a bit more general solution than, however it would require passing index to `Layers.Transformer.block`, which implies there is a single list of blocks. That is almost always the case, but Albert has actually N groups of M blocks, so there index would be ambiguous. So for now I went with another approach, which is to special case layer norm based on the name.

PR #436, opened 2025-12-28
Add Gemma 3 Text support

nyo16:
Adds support for Gemma 3 architecture, enabling FunctionGemma (`google/functiongemma-270m-it`) and other Gemma 3 models to run in Bumblebee.

   ### Why FunctionGemma?

   - **Lightweight** - Only 270M parameters, runs on CPU or modest GPU
   - **Function calling** - Specifically trained for tool/function invocation
   - **Easy to fine-tune** - Small enough to train on Google Colab T4
   - **Edge/IoT ready** - Perfect for home assistants, voice interfaces, embedded systems

   ### Gemma 3 Architecture Changes

   Gemma 3 has several key differences from Gemma v1:

   | Feature | Gemma v1 | Gemma 3 |
   |---------|----------|---------|
   | QK-norm | No | Yes (RMS norm on Q/K after projection) |
   | FFN layer norms | 1 (post-attention) | 3 (post-attention, pre-FFN, post-FFN) |
   | Residual order | Before post-attention norm | After post-attention norm |
   | Attention | Global only | Alternating local/global (5:1 ratio) |
   | RMS norm formula | `weight * normalized` | `(1 + weight) * normalized` |

   ### Files Changed

   - **`lib/bumblebee/text/gemma3.ex`** - Full Gemma 3 model implementation with custom decoder supporting QK-norm and extra layer norms
   - **`lib/bumblebee.ex`** - Model and tokenizer registrations for `Gemma3Model`, `Gemma3ForCausalLM`, etc.
   - **`lib/bumblebee/layers/transformer.ex`** - Per-layer `attention_window_size` callback for alternating local/global attention
   - **`test/bumblebee/text/gemma3_test.exs`** - Unit tests (require tiny-random models on HuggingFace)
   - **`notebooks/function_calling.livemd`** - Comprehensive Livebook example with:
     - `FunctionGemma.Schema` - Build function declarations
     - `FunctionGemma.Parser` - Parse function call responses
     - `FunctionGemma.Executor` - Execute parsed calls
     - `SmartHome` - Mock functions demo (lights, thermostat, weather)

   ### Example Usage

   ```elixir
   {:ok, model_info} = Bumblebee.load_model({:hf, "google/functiongemma-270m-it", auth_token: token})
   {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "google/functiongemma-270m-it", auth_token: token})
   {:ok, generation_config} = Bumblebee.load_generation_config({:hf, "google/functiongemma-270m-it", auth_token: token})

   serving = Bumblebee.Text.generation(model_info, tokenizer, generation_config,
     compile: [batch_size: 1, sequence_length: 512],
   )

   prompt = """
   <start_of_turn>developer
   You are a helpful assistant.
   <start_function_declaration>declaration:get_weather{description:<escape>Get
   weather<escape>,parameters:{properties:{location:{type:<escape>STRING<escape>}},required:[<escape>location<escape>],type:<escape>OBJECT<escape>}}<end_function_declaration><end_of_turn>
   <start_of_turn>user
   What's the weather in Paris?<end_of_turn>
   <start_of_turn>model
   """

   %{results: [%{text: text}]} = Nx.Serving.run(serving, prompt)
   # => "<start_function_call>call:get_weather{location:<escape>Paris<escape>}<end_function_call>"
   ```

   ## Test Plan

   - [x] Model loads without unused params warnings
   - [x] FunctionGemma generates correct function call format
   - [x] Multiple function types work (weather, lights, thermostat)
   - [x] Livebook runs end-to-end with mock function execution
   - [x] Unit tests pass (requires creating tiny-random-Gemma3* models on HuggingFace)"

### DIFF
--- a/lib/bumblebee/text/gemma3_text.ex
+++ b/lib/bumblebee/text/gemma3_text.ex
@@ -95,8 +95,7 @@ defmodule Bumblebee.Text.Gemma3Text do
         default: nil,
         doc: """
         a list of layer types for each layer, where each element is either `:sliding_attention`
-        (local attention with sliding window) or `:full_attention` (global attention).
-        If not provided, will be computed from `sliding_window_pattern`.
+        (local attention with sliding window) or `:full_attention` (global attention)
         """
       ],
       tie_word_embeddings: [
@@ -533,7 +532,9 @@ defmodule Bumblebee.Text.Gemma3Text do
 
   # Generate layer_types from sliding_window_pattern (default 6)
   # Pattern: every Nth layer uses full attention, others use sliding attention
-  defp generate_layer_types(num_blocks, sliding_window_pattern \\ 6) do
+  defp generate_layer_types(num_blocks) do
+    sliding_window_pattern = 6
+
     Enum.map(0..(num_blocks - 1), fn i ->
       if rem(i + 1, sliding_window_pattern) == 0 do
         :full_attention
@@ -576,16 +577,6 @@ defmodule Bumblebee.Text.Gemma3Text do
           end)
         end)
 
-      layer_types_converter = fn _name, value ->
-        types =
-          Enum.map(value, fn
-            "sliding_attention" -> :sliding_attention
-            "full_attention" -> :full_attention
-          end)
-
-        {:ok, types}
-      end
-
       opts =
         convert!(data,
           vocab_size: {"vocab_size", number()},
@@ -606,7 +597,14 @@ defmodule Bumblebee.Text.Gemma3Text do
           initializer_scale: {"initializer_range", number()},
           layer_norm_epsilon: {"rms_norm_eps", number()},
           sliding_window: {"sliding_window", optional(number())},
-          layer_types: {"layer_types", layer_types_converter},
+          layer_types:
+            {"layer_types",
+             list(
+               mapping(%{
+                 "sliding_attention" => :sliding_attention,
+                 "full_attention" => :full_attention
+               })
+             )},
           tie_word_embeddings: {"tie_word_embeddings", boolean()}
         ) ++ Shared.common_options_from_transformers(data, spec)
 
--- a/mix.exs
+++ b/mix.exs
@@ -64,6 +64,7 @@ defmodule Bumblebee.MixProject do
         "notebooks/llms.livemd",
         "notebooks/llms_rag.livemd",
         "notebooks/qwen3.livemd",
+        "notebooks/function_calling.livemd",
         "notebooks/fine_tuning.livemd",
         "examples/phoenix/README.md"
       ],
--- a/notebooks/function_calling.livemd
+++ b/notebooks/function_calling.livemd
@@ -1,4 +1,4 @@
-# Function Calling with FunctionGemma
+# Function calling with FunctionGemma
 
 ```elixir
 Mix.install([
@@ -8,36 +8,12 @@ Mix.install([
   {:kino, "~> 0.14.0"}
 ])
 
-Nx.global_default_backend({EXLA.Backend, client: :host})
+Nx.global_default_backend(EXLA.Backend)
 ```
 
 ## Why FunctionGemma?
 
-[FunctionGemma](https://huggingface.co/google/functiongemma-270m-it) is a compact 270M parameter model from Google, specifically designed for function calling tasks. Here's why it's a great choice:
-
-### Lightweight & Fast
-
-- **Only 270M parameters** - runs efficiently on CPU or modest GPU hardware
-- **Quick inference** - ideal for real-time applications
-- **Low memory footprint** - can run alongside other models
-
-### Perfect for Edge & IoT
-
-- **Home Assistants** - control smart home devices with natural language
-- **Voice Interfaces** - parse voice commands into structured function calls
-- **Embedded Systems** - fits on devices with limited resources
-
-### Easy to Fine-tune
-
-- **Train on Google Colab T4** - the small size makes fine-tuning accessible
-- **Custom function schemas** - adapt to your specific API or tool set
-- **Fast iteration** - experiment and deploy quickly
-
-### Structured Output
-
-- **Reliable function call format** - consistent `<start_function_call>` / `<end_function_call>` markers
-- **Easy to parse** - extract function name and arguments programmatically
-- **Multi-turn support** - handles function responses and follow-up calls
+[FunctionGemma](https://huggingface.co/google/functiongemma-270m-it) is a compact 270M parameter model from Google, specifically designed for function calling tasks.
 
 ## Loading the Model
 
@@ -531,18 +507,6 @@ Kino.Shorts.data_table(
 IO.puts("=== Final Smart Home State ===")
 SmartHome.get_status() |> IO.inspect(pretty: true)
 ```
-
-## Use Cases
-
-Here are some practical applications for FunctionGemma:
-
-| Use Case | Example Functions |
-|----------|-------------------|
-| **Smart Home** | control_light, set_thermostat, lock_door, play_music |
-| **Calendar** | create_event, list_events, reschedule_meeting |
-| **E-commerce** | search_products, add_to_cart, check_order_status |
-| **Database** | query_users, update_record, generate_report |
-| **DevOps** | deploy_service, check_status, scale_instances |
 
 ## Next Steps
 


test/bumblebee/text/gemma3_text_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    assert %Bumblebee.Text.Gemma3{architecture: :base} = spec
+
+    inputs = %{
+      "input_ids" => Nx.tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]]),
+      "attention_mask" => Nx.tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])
+    }
+
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 32}
```
jonatanklosko, line 21: Currently the tests only assert shape. They should assert output against reference values from Python transformers, as we do for all other models. That's how we verify our implementation behaves correctly.

lib/bumblebee/text/gemma3.ex
```diff
@@ -0,0 +1,726 @@
...
+            if rem(idx + 1, spec.global_attention_layer_interval) == 0 do
+              # Global attention (no window)
+              nil
+            else
+              # Local attention with sliding window
+              {spec.sliding_window, spec.sliding_window}
+            end
+
+          {hidden_state, attention, block_cache} =
+            gemma3_block(state.hidden_state,
```
jonatanklosko: Is there anything preventing us from using `Layers.Transformer.blocks`? If so, perhaps we can add extra options to `Layers.Transformer.blocks` to accommodate the necessary behaviour. We do that for all other models, so that the core transformer implementation is shared, rather than duplicated everywhere.

lib/bumblebee.ex
```diff
@@ -137,6 +137,13 @@ defmodule Bumblebee do
     "GemmaModel" => {Bumblebee.Text.Gemma, :base},
     "GemmaForCausalLM" => {Bumblebee.Text.Gemma, :for_causal_language_modeling},
     "GemmaForSequenceClassification" => {Bumblebee.Text.Gemma, :for_sequence_classification},
+    "Gemma3Model" => {Bumblebee.Text.Gemma3, :base},
+    "Gemma3ForCausalLM" => {Bumblebee.Text.Gemma3, :for_causal_language_modeling},
+    "Gemma3ForSequenceClassification" => {Bumblebee.Text.Gemma3, :for_sequence_classification},
+    "Gemma3TextModel" => {Bumblebee.Text.Gemma3, :base},
+    "Gemma3TextForCausalLM" => {Bumblebee.Text.Gemma3, :for_causal_language_modeling},
+    "Gemma3TextForSequenceClassification" =>
+      {Bumblebee.Text.Gemma3, :for_sequence_classification},
```
jonatanklosko: So there are actually two separate Gemma3 models. There is the `Gemma3TextModel` (and it's varieties), which is the base LLM. Then there is `Gemma3Model` (and it's varieties), which is **composed** of a text model (Gemma3TextModel) and a vision model, see the implementation [here](https://github.com/huggingface/transformers/blob/v5.0.0rc1/src/transformers/models/gemma3/modeling_gemma3.py#L803-L815). If you look at the [configuration file](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma3/configuration_gemma3.py), there is `Gemma3TextConfig` for the gemma text models and Gemma3Config for the multimodal model.

Now, the way this maps to Bumblebee is that we should have `Bumblebee.Text.Gemma3Text` module with options from `Gemma3TextConfig` and architectures: `:base` (= `Gemma3TextModel`), `:for_causal_language_modeling` (= `Gemma3ForCausalLM`), `:for_sequence_classification` (= `Gemma3TextForSequenceClassification`). The mapping should be this:

```suggestion
    "Gemma3TextModel" => {Bumblebee.Text.Gemma3Text, :base},
    "Gemma3TextForCausalLM" => {Bumblebee.Text.Gemma3Text, :for_causal_language_modeling},
    "Gemma3TextForSequenceClassification" => {Bumblebee.Text.Gemma3Text, :for_sequence_classification},
```

This PR effectively implements the text model, so you should rename the module to `Bumblebee.Text.Gemma3Text`. This should work as a LLM on its own.

---

Then, if we wanted to implement the multimodal Gemma3 models, it would be a separate module `Bumblebee.Multimodal.Gemma3`, with the corresponding config and composed models. We would also need to support whatever vision model the official Gemma3 checkpoints use. Note that we currently don't have a serving for text+image input, so to use the multimodal Gemma3, we would need a new serving type with all the necessary processing. Those changes are definitely much more involved, which I also mentioned in https://github.com/elixir-nx/bumblebee/pull/431#pullrequestreview-3615700435. Either way, those would be a few separate PRs :)

jonatanklosko:
Btw. I uploaded the tiny-random models to [bumblebee-testing](https://huggingface.co/bumblebee-testing/models?search=gemma3), so you can switch the tests to use those :)

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,597 @@
...
+      opts =
+        convert!(data,
+          vocab_size: {"vocab_size", number()},
+          max_positions: {"max_position_embeddings", number()},
+          hidden_size: {"hidden_size", number()},
+          num_blocks: {"num_hidden_layers", number()},
+          num_attention_heads: {"num_attention_heads", number()},
+          num_key_value_heads: {"num_key_value_heads", number()},
+          attention_head_size: {"head_dim", number()},
+          attention_scale_base: {"query_pre_attn_scalar", optional(number())},
```
jonatanklosko: Looking at the [config](https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/models/gemma3/configuration_gemma3.py#L88-L89) in transforemrs, this should never be `nil`, it has a default value though, which we should also have.

```suggestion
          attention_scale_base: {"query_pre_attn_scalar", number()},
```

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,597 @@
...
+      ],
+      attention_head_size: [
+        default: 256,
+        doc: "the size of the key, value, and query projection per attention head"
+      ],
+      attention_scale_base: [
+        default: nil,
+        doc: """
+        base value for computing attention scale. The attention scale is computed as
+        `attention_scale_base ** -0.5`. When `nil`, defaults to `:attention_head_size`
```
jonatanklosko:
```suggestion
        default: 256,
        doc: """
        base value for computing attention scale. The attention scale is computed as
        `attention_scale_base ** -0.5`.
```

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,597 @@
...
+          name: &2,
+          epsilon: spec.layer_norm_epsilon,
+          upcast: :all
+        ),
+      ffn:
+        &gated_ffn(&1, spec.intermediate_size, spec.hidden_size,
+          name: &2,
+          activation: spec.activation
+        ),
+      block_type: block_type,
```
jonatanklosko: Btw you can do this:

```suggestion
      block_type: &gemma3_block_impl(&1, &2, &3, spec),
```

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,597 @@
...
+    end
+
+    # Custom block_type function for Gemma 3's unique block structure
+    block_type = fn hidden_state, steps, block_name ->
+      gemma3_block_impl(hidden_state, steps, block_name, spec)
+    end
+
+    # Compute attention scale from attention_scale_base (defaults to attention_head_size)
+    attention_scale_base = spec.attention_scale_base || spec.attention_head_size
+    attention_scale = :math.pow(attention_scale_base, -0.5)
```
jonatanklosko:
```suggestion
    attention_scale = :math.pow(spec.attention_scale_base, -0.5)
```

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,595 @@
...
+        doc: "the sliding window size for local attention layers"
+      ],
+      global_attention_layer_interval: [
+        default: 6,
+        doc: """
+        the interval for global attention layers. In Gemma 3, every Nth layer uses global
+        attention while others use local (sliding window) attention. A value of 6 means
+        layers 5, 11, 17, 23... use global attention (5:1 local/global ratio)
+        """
+      ],
```
jonatanklosko: We never load this from the HuggingFace config, so it will cause discrepancies (could be the reason why tests failed).

In fact, the "interval" approach is deprecated, see [this code](https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/models/gemma3/configuration_gemma3.py#L188-L195).

I think we can go with option `:layer_types`, which is a list of either `:sliding_attention` or `:full_attention`. We have a similar config here:

https://github.com/elixir-nx/bumblebee/blob/a8caabddf6c83c2a81c4a76e16ea057724e4c956/lib/bumblebee/diffusion/controlnet.ex#L414-L422

We should also handle `"sliding_window_pattern"` if present. One way to do it would be to do something like this:

```elixir
# Support sliding_window_pattern for backward compatibility, see https://github.com/huggingface/transformers/blob/v5.0.0rc1/src/transformers/models/gemma3/configuration_gemma3.py#L188-L195
data =
  Map.put_new_lazy(data, "layer_types", fn ->
    pattern = data["sliding_window_pattern"] || 6
    # generate a list of Python-like layer types
  end)
```

jonatanklosko: Btw. I pushed updated checkpoints to HF with:

```
"layer_types": [
  "sliding_attention",
  "full_attention"
],
```

This way we can test both layer types.

You will need to update the reference values.

jonatanklosko:
> Note: There are still some small numerical differences between Elixir and Python outputs (max ~0.15), so I used slightly higher tolerances in the tests. This might be worth investigating further if needed, but the model works correctly with FunctionGemma examples.

We should be within 4 digit precision, or in other words we should not need `:atol` in the tests. I dropped a comment on one of the reasons the implementations don't match, but if the numbers still don't match, there may be something else too.

FWIW. I generated the checkpoints using this config:

```python
from transformers import Gemma3TextConfig, Gemma3TextModel, Gemma3ForCausalLM, Gemma3TextForSequenceClassification

config = Gemma3TextConfig(
  vocab_size=1024,
  hidden_size=32,
  num_hidden_layers=2,
  num_attention_heads=4,
  num_key_value_heads=2,
  intermediate_size=37,
  hidden_act="gelu_pytorch_tanh",
  hidden_dropout_prob=0.1,
  max_position_embeddings=512,
  type_vocab_size=16,
  is_decoder=False,
  initializer_range=0.02,
  pad_token_id=0,
  bos_token_id=2,
  eos_token_id=1,
  sliding_window=128,
  query_pre_attn_scalar=8,
  sliding_window_pattern=2
  # layer_types=[
  #   "sliding_attention",
  #   "full_attention"
  # ]
)

for c in [Gemma3TextModel, Gemma3ForCausalLM, Gemma3TextForSequenceClassification]:
  name = c.__name__
  c(config).save_pretrained(f"bumblebee-testing/tiny-random-{name}", repo_id=f"bumblebee-testing/tiny-random-{name}", push_to_hub=True)
```

nyo16:
Also the python code that i generated the test numbers 
```python
#!/usr/bin/env python3
  """Generate reference values for Bumblebee Gemma3Text tests."""

  import torch
  from transformers import AutoModel, AutoModelForCausalLM, AutoModelForSequenceClassification

  input_ids = torch.tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]])
  attention_mask = torch.tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])

  # :base test
  model = AutoModel.from_pretrained("bumblebee-testing/tiny-random-Gemma3TextModel")
  model.eval()
  with torch.no_grad():
      outputs = model(input_ids, attention_mask=attention_mask)
      print(":base - hidden_state[0, 1:4, 1:4]:")
      print(outputs.last_hidden_state[0, 1:4, 1:4])

  # :for_causal_language_modeling test
  model = AutoModelForCausalLM.from_pretrained("bumblebee-testing/tiny-random-Gemma3ForCausalLM")
  model.eval()
  with torch.no_grad():
      outputs = model(input_ids, attention_mask=attention_mask)
      print("\n:for_causal_language_modeling - logits[0, 1:4, 1:4]:")
      print(outputs.logits[0, 1:4, 1:4])

  # :for_sequence_classification test
  model = AutoModelForSequenceClassification.from_pretrained(
      "bumblebee-testing/tiny-random-Gemma3TextForSequenceClassification"
  )
  model.eval()
  with torch.no_grad():
      outputs = model(input_ids, attention_mask=attention_mask)
      print("\n:for_sequence_classification - logits:")
      print(outputs.logits)
```

lib/bumblebee/text/gemma3_text.ex
```diff
@@ -0,0 +1,648 @@
...
+          num_blocks: {"num_hidden_layers", number()},
+          num_attention_heads: {"num_attention_heads", number()},
+          num_key_value_heads: {"num_key_value_heads", number()},
+          attention_head_size: {"head_dim", number()},
+          attention_scale_base: {"query_pre_attn_scalar", number()},
+          intermediate_size: {"intermediate_size", number()},
+          activation: {"hidden_activation", activation()},
+          use_attention_bias: {"attention_bias", boolean()},
+          rotary_embedding_base: {"rope_theta", number()},
+          rotary_embedding_base_local: {"rope_local_base_freq", number()},
```
jonatanklosko: `rope_local_base_freq` does not actually exit on the config. Looks like now they have a whole configuration struct for rope, and in this case one per layer type. I need to investigate more how to handle this with best compatibility.

jonatanklosko, line 594: `rope_local_base_freq` does not actually exit on the config. Looks like now they have a whole configuration struct for rope, and in this case one per layer type. I will look into handling the new config shape.

jonatanklosko: Actually, `rope_local_base_freq` is still present in the config on Transformers v4, and v5 is still a release candidate. I pushed a small update, but I think we can merge this, I will make the rope config update separately.

nyo16: Awesome! Thank you @jonatanklosko

PR #438, opened 2026-01-03
feat: Add SigLIP model support

georgeguimaraes:
Adds support for SigLIP (Sigmoid Loss for Language Image Pre-Training), a CLIP-like model that uses sigmoid loss instead of contrastive loss.

The implementation includes:

- `Bumblebee.Multimodal.SigLip` for the combined text-image model
- `Bumblebee.Text.SigLipText` for the text encoder
- `Bumblebee.Vision.SigLipVision` for the vision encoder (with `:base` and `:for_image_classification` architectures)

SigLIP differs from CLIP mainly in the attention pooling head for the vision model and the addition of a learnable bias to the logits. The vision model uses a multi-head attention pooling mechanism with a learned probe vector.

Regarding naming: I went with `SigLip` (capitalizing each word part) to follow the pattern used by `MpNet`.

Note on the test: the tiny model used for testing comes from `katuni4ka/tiny-random-SiglipModel` rather than the usual `hf-internal-testing` namespace. I couldn't find a tiny SigLIP model there. The model works correctly against real HuggingFace models like `google/siglip-base-patch16-224`.

I also started working on SigLIP2 support but left it out of this PR to keep things focused. The architecture is similar but has some differences in the vision encoder (uses SwiGLU activation, RMSNorm, and 2D RoPE).


lib/bumblebee/vision/siglip_vision.ex
```diff
@@ -0,0 +1,438 @@
...
+    name = opts[:name]
+
+    probe =
+      Layers.learned_embeddings(1, spec.hidden_size,
+        name: join(name, "probe"),
+        initializer: Axon.Initializers.normal()
+      )
+
+    attended =
+      multihead_attention(probe, hidden_state, spec, name: join(name, "attention"))
```
jonatanklosko, line 237: We can likely reuse `Bumblebee.Layers.Transformer.multi_head_attention/4`:

https://github.com/elixir-nx/bumblebee/blob/f519f9b04699cb29dca427f092ec9ce5088e727e/lib/bumblebee/layers/transformer.ex#L749

lib/bumblebee/vision/siglip_vision.ex
```diff
@@ -0,0 +1,438 @@
...
+            end
+          },
+          "bias" => {
+            [{"vision_model.head.attention", "in_proj_bias"}],
+            fn [bias] ->
+              chunk_size = div(Nx.axis_size(bias, 0), 3)
+              Nx.slice_along_axis(bias, 0, chunk_size, axis: 0)
+            end
+          }
+        },
```
jonatanklosko, line 395: We can define a slicing function parameterized by chunk name, like we do here:

https://github.com/elixir-nx/bumblebee/blob/f519f9b04699cb29dca427f092ec9ce5088e727e/lib/bumblebee/text/nomic_bert.ex#L326-L355

lib/bumblebee/multimodal/siglip.ex
```diff
@@ -0,0 +1,250 @@
...
+    Layers.output(%{
+      logits_per_text: logits_per_text,
+      logits_per_image: logits_per_image,
+      text_embedding: text_embedding,
+      image_embedding: image_embedding
+    })
+  end
+
+  defp normalize(x) do
+    Nx.divide(x, Nx.LinAlg.norm(x, ord: 2, axes: [-1], keep_axes: true))
```
jonatanklosko, line 162: Can use this:

https://github.com/elixir-nx/bumblebee/blob/f519f9b04699cb29dca427f092ec9ce5088e727e/lib/bumblebee/utils/nx.ex#L326

lib/bumblebee/multimodal/siglip.ex
```diff
@@ -0,0 +1,250 @@
...
+    logits_per_text =
+      similarity
+      |> scale_layer(
+        name: "logit_scale",
+        initializer: Axon.Initializers.full(spec.logit_scale_initial_value)
+      )
+      |> bias_layer(
+        name: "logit_bias",
+        initializer: Axon.Initializers.full(spec.logit_bias_initial_value)
+      )
```
jonatanklosko, line 149: There is `Axon.bias`, which should be the same?

lib/bumblebee/multimodal/siglip.ex
```diff
@@ -0,0 +1,250 @@
...
+      text_embedding: text_embedding,
+      image_embedding: image_embedding
+    })
+  end
+
+  defp normalize(x) do
+    Nx.divide(x, Nx.LinAlg.norm(x, ord: 2, axes: [-1], keep_axes: true))
+  end
+
+  defp scale_layer(input, opts) do
```
jonatanklosko, line 165: We have this in CLIP too:

https://github.com/elixir-nx/bumblebee/blob/f519f9b04699cb29dca427f092ec9ce5088e727e/lib/bumblebee/multimodal/clip.ex#L155-L171

So it would make sense to share this. But we already have a `scale` layer that multiplies:

https://github.com/elixir-nx/bumblebee/blob/f519f9b04699cb29dca427f092ec9ce5088e727e/lib/bumblebee/layers.ex#L507

What I would do is add an option to `Layers.scale` called `:transform`, and it would accept an Nx function to apply to the scale itself. Then here and in CLIP we would pass `transform: &Nx.exp/1`.

lib/bumblebee.ex
```diff
@@ -116,6 +116,9 @@ defmodule Bumblebee do
     "CLIPModel" => {Bumblebee.Multimodal.Clip, :base},
     "CLIPTextModel" => {Bumblebee.Text.ClipText, :base},
     "CLIPVisionModel" => {Bumblebee.Vision.ClipVision, :base},
+    "SiglipModel" => {Bumblebee.Multimodal.SigLip, :base},
+    "SiglipTextModel" => {Bumblebee.Text.SigLipText, :base},
+    "SiglipVisionModel" => {Bumblebee.Vision.SigLipVision, :base},
```
jonatanklosko, line 121: We are missing this mapping:

```suggestion
    "SiglipModel" => {Bumblebee.Multimodal.SigLip, :base},
    "SiglipTextModel" => {Bumblebee.Text.SigLipText, :base},
    "SiglipVisionModel" => {Bumblebee.Vision.SigLipVision, :base},
    "SiglipForImageClassification" => {Bumblebee.Vision.SigLipVision, :for_image_classification},
```

lib/bumblebee.ex
```diff
@@ -229,6 +232,8 @@ defmodule Bumblebee do
 
   @transformers_class_to_featurizer %{
     "CLIPFeatureExtractor" => Bumblebee.Vision.ClipFeaturizer,
+    "SiglipImageProcessor" => Bumblebee.Vision.ClipFeaturizer,
```
jonatanklosko, line 235: `SiglipImageProcessor` is different from CLIP (e.g. it doesn't center crop). It should be it's own featurizer

lib/bumblebee.ex
```diff
@@ -229,6 +232,8 @@ defmodule Bumblebee do
...
+    "Siglip2ImageProcessor" => Bumblebee.Vision.ClipFeaturizer,
```
jonatanklosko, line 236: Siglip2 is a separate altogether.

```suggestion
```

test/bumblebee/multimodal/siglip_test.exs
```diff
@@ -0,0 +1,45 @@
+defmodule Bumblebee.Multimodal.SigLipTest do
+  use ExUnit.Case, async: true
+
+  import Bumblebee.TestHelpers
+
+  @moduletag model_test_tags()
+
+  test ":base" do
+    assert {:ok, %{model: model, params: params, spec: spec}} =
+             Bumblebee.load_model({:hf, "katuni4ka/tiny-random-SiglipModel"})
```
jonatanklosko, line 10: hf-internal-testing has repos for [all four models](https://huggingface.co/hf-internal-testing/models?search=siglip), so we should test against that. Note that we need tests for the text and vision models in separate files.

PR #440, opened 2026-01-03
Add Nomic BERT model

georgeguimaraes:
I needed nomic-embed-text-v1.5 for a project, so I added support for it.

It's a BERT variant with a few differences:
- Rotary position embeddings (base 1000)
- SwiGLU activation in FFN
- No biases in attention/FFN layers
- Combined Wqkv projection

I tested it against Python transformers and the outputs match within floating point precision (~2e-6).

```
Hello world: ✓ PASS (max diff: 2.0e-6)
The quick brown fox: ✓ PASS (max diff: 3.0e-6)
```

### DIFF
--- a/lib/bumblebee/text/nomic_bert.ex
+++ b/lib/bumblebee/text/nomic_bert.ex
@@ -180,30 +180,7 @@ defmodule Bumblebee.Text.NomicBert do
         name: "encoder"
       )
 
-    # Mean pooling over non-masked tokens
-    pooled_state =
-      Layers.if_present inputs["attention_mask"] do
-        Axon.layer(
-          fn hidden_state, attention_mask, _opts ->
-            # Expand mask for broadcasting with hidden_size
-            mask = Nx.new_axis(attention_mask, -1)
-            # Mask out padding tokens
-            masked = Nx.multiply(hidden_state, mask)
-            # Sum and normalize by actual sequence length
-            sum = Nx.sum(masked, axes: [1])
-            count = Nx.sum(mask, axes: [1])
-            Nx.divide(sum, Nx.max(count, 1.0e-9))
-          end,
-          [encoder_outputs.hidden_state, inputs["attention_mask"]]
-        )
-      else
-        Axon.layer(
-          fn hidden_state, _opts ->
-            Nx.mean(hidden_state, axes: [1])
-          end,
-          [encoder_outputs.hidden_state]
-        )
-      end
+    pooled_state = pooler(encoder_outputs.hidden_state, spec, name: "pooler")
 
     %{
       hidden_state: encoder_outputs.hidden_state,
@@ -264,6 +241,18 @@ defmodule Bumblebee.Text.NomicBert do
       output_use_bias: false,
       name: join(name, "blocks")
     )
+  end
+
+  defp pooler(hidden_state, spec, opts) do
+    name = opts[:name]
+
+    hidden_state
+    |> Layers.take_token(index: 0, axis: 1)
+    |> Axon.dense(spec.hidden_size,
+      kernel_initializer: kernel_initializer(spec),
+      name: join(name, "output")
+    )
+    |> Axon.tanh()
   end
 
   defp gated_ffn(hidden_state, intermediate_size, output_size, opts) do
@@ -342,7 +331,8 @@ defmodule Bumblebee.Text.NomicBert do
         "encoder.blocks.{n}.ffn.up" => "encoder.layers.{n}.mlp.fc11",
         "encoder.blocks.{n}.ffn.gate" => "encoder.layers.{n}.mlp.fc12",
         "encoder.blocks.{n}.ffn.down" => "encoder.layers.{n}.mlp.fc2",
-        "encoder.blocks.{n}.output_norm" => "encoder.layers.{n}.norm2"
+        "encoder.blocks.{n}.output_norm" => "encoder.layers.{n}.norm2",
+        "pooler.output" => "pooler.dense"
       }
     end
 
--- a/test/bumblebee/text/nomic_bert_test.exs
+++ b/test/bumblebee/text/nomic_bert_test.exs
@@ -30,7 +30,7 @@ defmodule Bumblebee.Text.NomicBertTest do
 
     assert_all_close(
       outputs.pooled_state[[.., 1..3]],
-      Nx.tensor([[0.5064, -0.6647, -0.9303]])
+      Nx.tensor([[0.0197, -0.2129, -0.0071]])
     )
   end
 end


lib/bumblebee/text/nomic_bert.ex
```diff
@@ -0,0 +1,410 @@
...
+        ) ++ Shared.common_options_from_transformers(data, spec)
+
+      # Nomic uses 4 * n_embd for intermediate size if not specified
+      opts =
+        if opts[:intermediate_size] do
+          opts
+        else
+          hidden_size = opts[:hidden_size] || spec.hidden_size
+          Keyword.put(opts, :intermediate_size, 4 * hidden_size)
+        end
```
jonatanklosko: Instead of doing this on loading, we should rather make `intermediate_size` optional (default `nil`) and then handle that in the implementation. This is what we do in GPT2:

https://github.com/elixir-nx/bumblebee/blob/2252d44a06ae9d9c3cbbef2987817afd4339be92/lib/bumblebee/text/gpt2.ex#L411

(the NomicBERT config is actually based on GPT2)

test/bumblebee/text/nomic_bert_test.exs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 6, 768}
+    assert Nx.shape(outputs.pooled_state) == {1, 768}
+
+    # Values verified against Python transformers
+    assert_all_close(
+      outputs.hidden_state[[.., 0, 0..4]],
```
jonatanklosko: We typically test `[[.., 1..3, 1..3]]` slice, so let's do it here too.

test/bumblebee/text/nomic_bert_test.exs
```diff
@@ -0,0 +1,42 @@
...
+      Nx.tensor([[1.3752, 0.7431, -4.6988, -0.6574, 2.1887]]),
+      atol: 1.0e-3
```
jonatanklosko: We should not need to override `:atol` here, if it fails with the default 1.0e-4, it most likely means the model implementation does not match the reference one, so that needs to be addressed.

lib/bumblebee/text/nomic_bert.ex
```diff
@@ -0,0 +1,410 @@
...
+
+  defp encoder(hidden_state, position_ids, attention_mask, attention_head_mask, spec, opts) do
+    name = opts[:name]
+
+    # Nomic BERT uses postnorm (like standard BERT):
+    # Each block:
+    #   attn_output = attention(hidden_states)
+    #   hidden_states = norm1(attn_output + hidden_states)
+    #   ffn_output = ffn(hidden_states)
+    #   hidden_states = norm2(ffn_output + hidden_states)
```
jonatanklosko: This seems standard, is there any reason we cannot use `Layers.Transformer.blocks`, as we do for most other models?

test/bumblebee/text/nomic_bert_test.exs
```diff
@@ -0,0 +1,42 @@
...
+    assert %Bumblebee.Text.NomicBert{architecture: :base} = spec
+    assert spec.hidden_size == 768
+    assert spec.num_blocks == 12
+    assert spec.num_attention_heads == 12
+    assert spec.rotary_embedding_base == 1000
+
+    inputs = %{
+      "input_ids" => Nx.tensor([[101, 2023, 2003, 1037, 3231, 102]]),
+      "attention_mask" => Nx.tensor([[1, 1, 1, 1, 1, 1]])
+    }
```
jonatanklosko, line 17: Nit: let's use the same value as we do in other tests. This way it's easier to test against hf/transformers without copying different input for each model.

```suggestion
    inputs = %{
      "input_ids" => Nx.tensor([[10, 20, 30, 40, 50, 60, 70, 80, 0, 0]]),
      "attention_mask" => Nx.tensor([[1, 1, 1, 1, 1, 1, 1, 1, 0, 0]])
    }
```

test/bumblebee/text/nomic_bert_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  use ExUnit.Case, async: true
+
+  import Bumblebee.TestHelpers
+
+  @moduletag model_test_tags()
+
+  @tag :slow
+  test ":base" do
+    assert {:ok, %{model: model, params: params, spec: spec}} =
+             Bumblebee.load_model({:hf, "nomic-ai/nomic-embed-text-v1.5"})
```
jonatanklosko: Once the tests pass as expected, I will upload a tiny random model.

test/bumblebee/text/nomic_bert_test.exs
jonatanklosko: It's worth having a look at [the config attributes](https://huggingface.co/nomic-ai/nomic-bert-2048/blob/main/configuration_hf_nomic_bert.py#L7-L31). For example
`mlp_fc1_bias` and `mlp_fc2_bias` seem relevant, but we don't import those.

georgeguimaraes: yeah, i was disabling bias for every model. Now I'm importing the appropriate config.

lib/bumblebee/text/nomic_bert.ex
georgeguimaraes: Good catch! I initially thought the model implementation should be self-contained, but honestly I just didn't pay close attention to what Layers.Transformer.blocks already provides. Thanks

jonatanklosko: > I initially thought the model implementation should be self-contained

That's what hf/transformers do, but it has tradeoffs, and in our case we ended up normalizing and sharing all of the core transformer logic. This makes it easier to maintain and add new models, because most LLMs introduce just a few differences.

test/bumblebee/text/nomic_bert_test.exs
jonatanklosko: Pushed to `bumblebee-testing/tiny-random-NomicBertModel`.

georgeguimaraes:
using tiny model now

test/bumblebee/text/nomic_bert_test.exs
```diff
@@ -0,0 +1,36 @@
...
+
+    outputs = Axon.predict(model, params, inputs)
+
+    assert Nx.shape(outputs.hidden_state) == {1, 10, 32}
+    assert Nx.shape(outputs.pooled_state) == {1, 32}
+
+    assert_all_close(
+      outputs.hidden_state[[.., 1..3, 1..3]],
+      Nx.tensor([
+        [[1.5269, -0.3709, -0.6235], [0.0301, -0.1500, -1.0316], [-1.4733, -1.1167, 0.2346]]
```
jonatanklosko: I was double checking, and I noticed that the output from Python transformers is different on every call. Then I realised that the config had dropout probabilities enabled, as done for training. I updated the repo to provide deterministic output, and now I get these:

```
last_hidden_state:
tensor([[[ 0.3062, -2.1607, -0.1782],
         [ 0.5486,  0.5353, -0.0453],
         [ 0.6323, -0.7370, -2.0245]]], grad_fn=<SliceBackward0>)
pooler_output:
tensor([[ 0.0197, -0.2129, -0.0071]], grad_fn=<SliceBackward0>)
```

Let's update accordingly :)

georgeguimaraes: done!

jonatanklosko: The pooler output seems different, no?

georgeguimaraes: Done! The pooler values now match yours.

Interestingly, Python transformers doesn't load the pooler weights by default for NomicBertModel. The from_pretrained method sets add_pooling_layer=False for this class, so pooler_output returns None. The weights exist in safetensors but get filtered out during loading.

To reproduce your values, I had to manually load the pooler weights. Our Bumblebee implementation now has a proper pooler layer and loads the weights correctly.

jonatanklosko: Yeah, I was loading with:

```python
model = AutoModel.from_pretrained("bumblebee-testing/tiny-random-NomicBertModel", trust_remote_code=True, add_pooling_layer=True)
```

In the model implementation it seems that it [should be `True` by default](https://huggingface.co/nomic-ai/nomic-bert-2048/blob/main/modeling_hf_nomic_bert.py#L1869), perhaps this doesn't work with `AutoModel` for some reason.

PR #441, opened 2026-01-04
feat: Add :progress_bar_step config option

georgeguimaraes:
The progress bar updates on every network chunk during downloads. This floods CI logs because the escape codes for in-place updates don't work there, so each update ends up on a new line.

Added a `:progress_bar_step` config option that limits updates to percentage boundaries:

```elixir
# Update every 10%
config :bumblebee, :progress_bar_step, 10
```

Default is `nil` (current behavior). Setting it to `10` gives you ~10 lines of output instead of hundreds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,16 +42,6 @@ config :nx, default_backend: EXLA.Backend
 
 To use GPUs, you must [set the `XLA_TARGET` environment variable accordingly](https://github.com/elixir-nx/xla#usage).
 
-You can also configure the progress bar displayed during model downloads:
-
-```elixir
-# Update every 10% - useful for CI environments
-config :bumblebee, :progress_bar_step, 10
-
-# Disable progress bar entirely
-config :bumblebee, :progress_bar_enabled, false
-```
-
 In notebooks and scripts, use the following `Mix.install/2` call to both install and configure dependencies:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


lib/bumblebee/utils/http.ex
```diff
@@ -106,6 +108,28 @@ defmodule Bumblebee.Utils.HTTP do
...
 
+  defp maybe_render_progress(state, part_size)
+       when is_nil(state.total_size) or part_size == state.total_size do
+    state
+  end
+
+  defp maybe_render_progress(state, _part_size) do
+    if Bumblebee.Utils.progress_bar_enabled?() do
+      step = Bumblebee.Utils.progress_bar_step()
+      percent = trunc(state.size / state.total_size * 100)
```
jonatanklosko, line 118: Instead of storing `last_percent`, we can do `trunc((state.size - part_size) / state.total_size * 100)` to compute it here. I think it's preferable since it makes things more encapsulated.

lib/bumblebee/utils.ex
```diff
@@ -8,4 +8,15 @@ defmodule Bumblebee.Utils do
...
     Application.get_env(:bumblebee, :progress_bar_enabled, true)
   end
+
+  @doc """
+  Returns the progress bar update step in percent.
+
+  When set, progress updates only when crossing step boundaries (e.g., 10%, 20%, ...).
+  Defaults to `nil`, which updates on every chunk.
+  """
+  @spec progress_bar_step :: non_neg_integer() | nil
```
jonatanklosko, line 18: It may be reasonable to make the default `1` or `0.5` to limit the IO, it shouldn't make much difference UX wise. Arguably if we make it `1`, we don't necessarily need it to be configurable, since that should be fine on CI, but either works for me.

README.md
```diff
@@ -42,6 +42,16 @@ config :nx, default_backend: EXLA.Backend
...
 
+You can also configure the progress bar displayed during model downloads:
+
+```elixir
+# Update every 10% - useful for CI environments
+config :bumblebee, :progress_bar_step, 10
+
+# Disable progress bar entirely
+config :bumblebee, :progress_bar_enabled, false
+```
```
jonatanklosko: I wouldn't mention it that early in the readme, since it's rather specific. Also this part of README does not become the docs, but there is a chunk below that ends up in the docs. Let's add a `## Global configuration` section under `### Tokenizer support`.

lib/bumblebee/utils.ex
georgeguimaraes: default is now 1

lib/bumblebee/utils/http.ex
```diff
@@ -106,6 +108,27 @@ defmodule Bumblebee.Utils.HTTP do
     :ok
   end
 
+  defp maybe_render_progress(state, part_size)
+       when is_nil(state.total_size) or part_size == state.total_size do
```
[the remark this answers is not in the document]
georgeguimaraes: since progress_bar will give us a chunk, total_size will not be 0

jonatanklosko (approved):

PR #442, opened 2026-01-07
Add Qwen3-VL vision-language model support

nyo16:
## Summary

   This PR adds full support for the Qwen3-VL vision-language model family, enabling image-to-text generation with Bumblebee.

   **Model**: `Qwen/Qwen3-VL-2B-Instruct` (and other sizes)
   **Architecture**: `Qwen3VLForConditionalGeneration`

   ## Features

   ### Vision Encoder (`Bumblebee.Vision.Qwen3VLVision`)
   - 3D convolution patch embedding (supports video temporal dimension)
   - 2D spatial rotary position embeddings for accurate spatial understanding
   - Bilinear interpolation for position embeddings (handles variable image sizes)
   - Patch merger with spatial reduction (2x2 → 1)
   - **DeepStack feature extraction** from layers [5, 11, 17]

   ### Text Decoder
   - Based on Qwen3 architecture with QK-norm
   - Visual token substitution (replaces image placeholder tokens with vision embeddings)
   - **DeepStack injection** at decoder layers [0, 1, 2]
   - Full rotary position embedding support

   ### Featurizer (`Bumblebee.Vision.Qwen3VLFeaturizer`)
   - Image preprocessing with configurable resize
   - Automatic padding to patch-aligned dimensions
   - Support for both images and video frames
   - Outputs flattened patches: `{num_patches, channels * temporal * patch_h * patch_w}`

   ### DeepStack Implementation
   DeepStack provides multi-scale visual information by:
   1. Extracting hidden states from vision encoder layers [5, 11, 17] (1-indexed)
   2. Passing each through separate merger MLPs with **postshuffle norm** (norm AFTER spatial merge)
   3. Injecting features into text decoder at layers [0, 1, 2]
   4. Formula: `hidden_states[visual_mask] += deepstack_features[layer_idx]`

   ### Infrastructure Changes
   - Added `post_block_hook` option to `Layers.Transformer.blocks` for per-layer injection

   ## Files Changed

   ### New Files
   - `lib/bumblebee/multimodal/qwen3_vl.ex` - Main VL model
   - `lib/bumblebee/vision/qwen3_vl_vision.ex` - Vision encoder
   - `lib/bumblebee/vision/qwen3_vl_featurizer.ex` - Image preprocessing
   - `test/bumblebee/multimodal/qwen3_vl_test.exs` - Tests
   - `notebooks/qwen3_vl.livemd` - Usage examples

   ### Modified Files
   - `lib/bumblebee.ex` - Model/featurizer registrations
   - `lib/bumblebee/layers/transformer.ex` - Added `post_block_hook` option

   ## Test Results

   ```
   # Unit test with tiny model
   mix test test/bumblebee/multimodal/qwen3_vl_test.exs
   1 test, 0 failures

   # Real model test with image (448x448)
   Image: 448x448x3
   Patches: 784 → 196 visual tokens (after merge)
   Generated: "This image shows a close-up of a single, small, dark-colored object..."
   ✓ DeepStack injection test with real image PASSED!
   ```

   ## Usage Example

   ```elixir
   {:ok, model_info} = Bumblebee.load_model({:hf, "Qwen/Qwen3-VL-2B-Instruct"}, type: :bf16)
   {:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "Qwen/Qwen3-VL-2B-Instruct"})
   {:ok, featurizer} = Bumblebee.load_featurizer({:hf, "Qwen/Qwen3-VL-2B-Instruct"},
     module: Bumblebee.Vision.Qwen3VLFeaturizer)

   # Load and process image
   image = StbImage.read_file!("photo.jpg")
   image_inputs = Bumblebee.apply_featurizer(featurizer, image)

   # Build prompt with image placeholder
   prompt = """
   <|im_start|>user
   <|vision_start|><|image_pad|><|vision_end|>Describe this image.<|im_end|>
   <|im_start|>assistant
   """

   text_inputs = Bumblebee.apply_tokenizer(tokenizer, prompt)
   inputs = Map.merge(text_inputs, image_inputs)

   # Run inference
   outputs = Axon.predict(model_info.model, model_info.params, inputs)
   ```

   ## Parameter Loading

   All parameters load correctly with no warnings:
   - Vision encoder: patch_embed, pos_embed, blocks.{0-23}, merger
   - DeepStack mergers: deepstack_merger_list.{0-2} (9 params total)
   - Text decoder: embedder, decoder.blocks.{0-27}, output_norm, lm_head

   ## References

   - [Qwen3-VL on HuggingFace](https://huggingface.co/Qwen/Qwen3-VL-2B-Instruct)
   - [Qwen3-VL Technical Report](https://arxiv.org/abs/2409.12191)


fire:
What's the process to get something on to the review sequence?

notebooks/qwen3_vl.livemd
```diff
@@ -0,0 +1,127 @@
...
+"
+
+# Process image
+image_inputs = Bumblebee.apply_featurizer(featurizer, image)
+
+# Combine prompt with image inputs
+generation_input = %{
+  prompt: prompt,
+  images: image_inputs
+}
```
jonatanklosko: Note to self: this doesn't exit, we likely want a separate serving, I need to think about the API. Also note that the image_pad token needs to be repeated as part of preprocessing to match the number of image patches.

petermueller:
I have some improvements on top of this branch that I'm fairly certain get it to parity with the python implementation. 
But I was helped extensively by Claude and although I understand a lot of it, I am still working through a lot of things I don't understand to be able to have a useful and intelligent conversation that doesn't waste reviewers time. I'm trying to match a lot of the terminology with my understanding from going through Sean Moriarty's book

@nyo16 or (@polvalente or @jonatanklosko?), would it be useful for me to open a PR to view or compare things?

I don't want to waste people's time asking for review of something that I have a grasp of but not an intimate knowledge of yet. So if it's not something you're actively interested in looking at, I don't want to ask for your time, but just generally want to know if I should open a PR. Or potentially open a PR on @nyo16's fork. I've generally got it down to where I think any discrepancies between it and python are due to point floating point math differences or just general non-determinism

jonatanklosko:
Sorry for the late reply.

I think the main challenge here is about static shapes. We need to compile the model with specific tensor sizes, but it should support variable number of images images of various sizes (like the Python version), up to an upper bound that we compile for. For plain text models that accept token sequence and attention mask, we compile for a fixed size (upper limit) and at runtime we pad each input to match that size. We need to do something similar to support variable number of images of different sizes (the question is if we can implement the model/graph so that those can be dynamic). The images are flattened into patches, so that's good, but from a brief look the Python implementation also relies on image dimensions information (`image_grid_thw`), which I think this PR is missing (I may be wrong).

@petermueller if you manage to get Python parity, handling the `image_grid_thw` parts, a PR would definitely be interesting :)

nyo16:
@jonatanklosko you are correct, I can take a look later this week to add the dimensions and i think python has also "profiles" like low, mid, high quality.

fire:
Is it required to have to have the variable number shapes or can we do that on a followup pull request?

jonatanklosko:
> Is it required to have to have the variable number shapes or can we do that on a followup pull request?

I think solving this is required for any practical use. Also, without adapting the implementation to deal with image sizes like the Python version, the model likely doesn't work as expected.

petermueller:
I have parity-level fixes on top of this PR
at https://github.com/petermueller/bumblebee/tree/feat/qwen3-vl
- top-10 next-token agreement 10/10 vs HF, last-position logit max diff ~0.05. 
- `image_grid_thw` is wired through featurizer -> position_ids -> vision encoder

 
 Static shapes (compile once + pad to upper bound) is still missing. Currently the model rebuilds per-grid. I can work on that this weekend though. How would people like things threaded. PR to yours, @nyo16 or some other way?

petermueller:
> without adapting the implementation to deal with image sizes like the Python version, the model likely doesn't work as expected.

I can confirm this part. I got a lot of "there's five cats on top of each other" while adding things 😅

nyo16:
@petermueller feel free to add anything you need to this PR.

nyo16:
Quick update — pushed three commits to address the open feedback. TL;DR: image_grid_thw is now threaded through end-to-end, the model gets coherent output on a real Qwen3-VL-2B-Instruct run, and there's a
  compile/run path that addresses the static-shapes concern by padding to upper bounds.

  Commit 1 (7c19bb0) — grid_thw + smart-resize + multi-image
  - Featurizer implements smart-resize (mirrors qwen-vl-utils): preserves aspect ratio, snaps to patch_size * merge_size, clamps total pixels between :min_pixels and :max_pixels. Added a :quality preset (:low /
  :medium / :high) as a convenience over the raw caps.
  - Featurizer accepts a list of images of different sizes in one call. Patches are emitted in windowed order (every 4 consecutive patches form one 2×2 merge block), which makes the patch merger reshape
  shape-agnostic — that's the trick that lets multi-image work without per-image rebuilds.
  - Vision encoder: image_grid_thw is a real model input now. Per-patch (row, col, grid_h, grid_w, image_id) are derived from grid_thw via Nx ops and drive both bilinear pos-embed interpolation and 2D RoPE. The
  attention mask is block-diagonal per image so patches from different images cannot cross-attend.
  - 13 new tests, full fast suite green (285 passed, 0 regressions).

  Commit 2 (bbb8dd6) — multimodal generation helper
  - Bumblebee.Multimodal.ImageTextToText.generate/6 does featurization + <|image_pad|> expansion + Text.Generation.build_generate in one call.
  - I tried to wrap this in an Nx.Serving first, but Nx.Batch.concatenate requires every tensor to share the same first-axis size — that breaks for Qwen3-VL since pixel_values is {num_patches, _} while input_ids
  is {1, seq_len}. Function-based API is the practical interim.

  Commit 3 (1b149f7) — static-shape padding
  - Featurizer gets :max_patches and :max_num_images options. When set, pixel_values is zero-padded along the patches axis and image_grid_thw is padded with [0, 0, 0] rows.
  - Vision encoder derives a patch_valid mask (i < total_real_patches) from grid_thw and ANDs it into the attention mask. Padded patches therefore neither attend nor are attended to — their embedding
  contributions drop out of the output.
  - New compile/5 + run/3 API for compile-once-reuse. On a 640×480 COCO image, cold = 27.3s, warm = 10.1s, 2.7x speedup with identical output (greedy decode).

  Validation against real model + HuggingFace reference
  Ran Qwen/Qwen3-VL-2B-Instruct on test/fixtures/images/coco/39769.jpeg:
  - Greedy decode output: "A group of cats are lying on a pink blanket with remote controls." — the image is two cats on a pink couch with TV remotes, so the model is reading it correctly.
  - Side-by-side with HuggingFace transformers 5.9.0 / PyTorch 2.12 / bf16, same prompt and min_pixels/max_pixels: same image_grid_thw [[1, 26, 36]], same pixel_values shape, same prompt length, top-3 identical
  and in same order, 8/10 top-10 next-token IDs shared. The two divergences are reasonable image-description starters at the rank floor where bf16 rounding can flip order.

  Coordination with @petermueller
  This took your fork's grid-threading insight (thank you — that was the unblocker) and pushed it through to image_grid_thw as a real runtime input plus multi-image support, rather than a spec-level
  with_image_grid rebuild. If you have parity fixes I missed, happy to layer them on or swap in.

  What's still left (genuine follow-ups, not in this PR):
  - 3D M-RoPE on the text decoder. Python uses temporal/height/width per-axis frequency schedules; we still use 1D RoPE. Wider parity gain available here.
  - A true Nx.Serving wrapper around compile/5. The function API works but doesn't plug into Bumblebee's serving infrastructure cleanly because of the mixed-axis-0 shape mismatch.
  - Video frame plumbing — featurizer already handles temporal > 1 but the prompt expander doesn't yet distinguish <|image_pad|> from <|video_pad|>.

  @jonatanklosko — I think this addresses both the image_grid_thw and "static shapes" comments. Happy to slice this differently if you want any of it split into separate PRs.

petermueller:
Reviewed your changes — runtime `image_grid_thw` is a much better
  base than what I had. Three things I noted that I can branch off your `feat/qwen3-vl` that I think are still useful:

  1. **DeepStack off-by-one** in `qwen3_vl_vision.ex:373` —
     `Enum.map(&(&1 - 1))` treats `deepstack_visual_indexes` as 1-indexed,
     but HF compares them against `enumerate(self.blocks)` (0-indexed). For
     `[5, 11, 17]` we end up grabbing layers 4/10/16 instead. In my f32
     reference run, the fix lifted top-10 next-token agreement 2/10 → 10/10.

  2. **mRoPE in the text decoder** — your follow-up item; I'll wire it in.
     Single-image first; multi-image mRoPE positions probably want their
     own design pass.

  3. **`mix bumblebee.compare_qwen3_vl`** — diffs per-block hidden states
     and attention intermediates vs HF PyTorch. Bringing it forward as
     ongoing validation; it's what surfaced (1) for me.
     
 Let me know if/how you'd want these. I think the Deepstack bug would be most useful in here, so I'm putting it on an updated branch from yours. The mix task I found really useful, but as it stands it relies on the mRoPE changes to be able to introspect things fully, so I'll put those separately.

nyo16:
@petermueller Good catch — pushed the DeepStack fix in 9e87345. The doc on :deepstack_visual_indexes claimed 1-indexed and the Enum.map(&(&1 - 1)) was acting on that wrong assumption.

  Re-ran both sides — HF Python reference re-executed and confirmed byte-identical to the previous run (top-10 IDs [28715, 1986, 785, 2124, 2082, 3862, 32, 11613, 95456, 641]), then ran Bumblebee bf16 on the same
   COCO image + prompt:

  Rank | HF (Python) | Bumblebee (before fix) | Bumblebee (after fix)
  -----|-------------|------------------------|----------------------
    1  | Based       | Based  ✓               | Based  ✓
    2  | This        | This   ✓               | This   ✓
    3  | The         | The    ✓               | The    ✓
    4  | Of          | An  (swapped)          | Of     ✓
    5  | An          | Of  (swapped)          | An     ✓
    6  | There       | A                      | There  ✓
    7  | A           | In                     | A      ✓
    8  | Two         | There                  | In
    9  | Certainly   | What                   | What
   10  | In          | Hello                  | Hello

  Top-7 now match HF exactly in order (was top-3 before). Still 8/10 on top-10 in bf16 — your 2→10 was on f32, which lines up with the remaining 2 being rank flips at the rounding noise floor. I haven't
  re-validated in f32 from this side; happy to if you want the number.

  On the other two:
  - mRoPE: go for it, separate branch makes sense. Single-image first is the right call — multi-image M-RoPE positions interact with how image_grid_thw indexes patches across the concatenated sequence, that
  deserves its own design pass.
  - mix bumblebee.compare_qwen3_vl: please bring it. Per-block hidden state diffs vs HF is exactly what's missing for this kind of work, and "wait for mRoPE first" is fine — keep it self-contained on the mRoPE
  branch.

  Thanks again for the review.

PR #444, opened 2026-01-11
feat: Add cross_encoder serving and fix text_classification token_type_ids

### LINKED ISSUE #251: Cross Encoder support
Hi, I'm currently trying to implement a feature called "hybrid search" inside of my application. It's based on returning query results from multiple databases and later scoring results together from multiple sources. To score them I want to use cross-encoder models ie. https://huggingface.co/cross-encoder/ms-marco-MiniLM-L-6-v2. 

I'm trying to understand if Bumblebee currently supports models like this and if so how to use it for that.

If it does I'll be very happy to write some documentation for that after getting some hints. If not it would be a very cool feature to handle those types of operations :)

georgeguimaraes:
I was experimenting with rerankers for [georgeguimaraes/arcana](https://github.com/georgeguimaraes/arcana) and found that cross-encoder models like `cross-encoder/ms-marco-MiniLM-L-6-v2` weren't producing correct scores.

The issue: `text_classification` was setting `return_token_type_ids: false`, which breaks sentence-pair inputs. Cross-encoders need token_type_ids to distinguish query tokens from document tokens. Without them, scores don't match Python's sentence-transformers.

Changes:
- Fixed `text_classification` to include token_type_ids (also added it to the compile template)
- Added a new `cross_encoder` serving with a cleaner API for the reranking use case

```elixir
serving = Bumblebee.Text.cross_encoder(model_info, tokenizer)
Nx.Serving.run(serving, {"query", "document"})
#=> %{score: 8.76}
```

The token_type_ids fix also benefits other sentence-pair tasks like NLI and entailment. If you don't want that change in text_classification, let me know.

Closes #251

### DIFF
--- a/lib/bumblebee/text.ex
+++ b/lib/bumblebee/text.ex
@@ -361,6 +361,68 @@ defmodule Bumblebee.Text do
   defdelegate text_classification(model_info, tokenizer, opts \\ []),
     to: Bumblebee.Text.TextClassification
 
+  @type cross_encoding_input :: {String.t(), String.t()}
+  @type cross_encoding_output :: %{score: number()}
+
+  @doc """
+  Builds serving for cross-encoder models.
+
+  Cross-encoders score text pairs by encoding them jointly through a
+  transformer with full cross-attention. This is commonly used for
+  reranking search results, semantic similarity, and natural language
+  inference tasks.
+
+  The serving accepts `t:cross_encoding_input/0` and returns
+  `t:cross_encoding_output/0`. A list of inputs is also supported.
+
+  ## Options
+
+    * `:compile` - compiles all computations for predefined input shapes
+      during serving initialization. Should be a keyword list with the
+      following keys:
+
+        * `:batch_size` - the maximum batch size of the input. Inputs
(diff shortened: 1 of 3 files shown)


lib/bumblebee/text.ex
```diff
@@ -361,6 +361,68 @@ defmodule Bumblebee.Text do
...
+      #=> %{score: 8.761}
+
+  """
+  @spec cross_encoder(
+          Bumblebee.model_info(),
+          Bumblebee.Tokenizer.t(),
+          keyword()
+        ) :: Nx.Serving.t()
+  defdelegate cross_encoder(model_info, tokenizer, opts \\ []),
+    to: Bumblebee.Text.CrossEncoder
```
jonatanklosko: Nit: let's rename to "cross encoding" to match all the other functions we already have (embeding, classifiction, etc):

```suggestion
  defdelegate cross_encoding(model_info, tokenizer, opts \\ []),
    to: Bumblebee.Text.CrossEncoding
```

georgeguimaraes: done

PR #447, opened 2026-01-28
fix: update URL to fetch Paul Graham essay

bendehaan:
The original URL is now a 404, replacing it with the new location.


jonatanklosko (approved):
Thanks :)

PR #453, opened 2026-04-10
Allow ffmpeg media path to contain SPACE

bougueil:
When building the ffmpeg command, ensures a path containing SPACE  is not sliced.


jonatanklosko (approved):
Thank you :)

PR #448, opened 2026-01-28
Add native FP8 model support. 

nyo16:
## Summary

  This PR adds support for loading and running FP8 (8-bit floating point) quantized models natively in Bumblebee. FP8 models use approximately half the memory
   of BF16 models while maintaining good inference quality.

  ## Changes

  ### Core FP8 Support
  - Add `preserve_source_types` option to `Bumblebee.load_model/2` to keep FP8 weights in their native format
  - Add `dequantize_kernel/3` function in `Bumblebee.Layers` for runtime FP8 → F32 conversion using scale_inv tensors
  - Update FP8 type detection to use the new nx/safetensors format `{:f8_e4m3fn, 8}`

  ### Qwen3 FP8 Integration
  - Add `params_mapping` for FP8 weight scales (`weight_scale_inv`) in Qwen3 architecture
  - Update transformer layers to apply dequantization when FP8 weights are detected

  ### Dependencies
  - Update to git versions of `nx`, `exla`, `torchx`, and `safetensors` for FP8 type support
  - Note: These can be switched back to hex versions once released

  ### Documentation
  - Add FP8 quantization section to the Qwen3 notebook with usage examples

  ## Usage

  ### Loading an FP8 Model

  ```elixir
  repo = {:hf, "Qwen/Qwen3-4B-Instruct-2507-FP8"}

  {:ok, model_info} = Bumblebee.load_model(repo,
    backend: EXLA.Backend,
    preserve_source_types: true
  )
  {:ok, tokenizer} = Bumblebee.load_tokenizer(repo)
  {:ok, generation_config} = Bumblebee.load_generation_config(repo)

  Running Inference

  generation_config =
    Bumblebee.configure(generation_config,
      max_new_tokens: 256,
      temperature: 0.7,
      strategy: %{type: :multinomial_sampling, top_p: 0.8, top_k: 20}
    )

  serving =
    Bumblebee.Text.generation(model_info, tokenizer, generation_config,
      compile: [batch_size: 1, sequence_length: 1024],
      defn_options: [compiler: EXLA]
    )

  # Using chat template format
  prompt = """
  <|im_start|>system
  You are a helpful assistant.<|im_end|>
  <|im_start|>user
  What are the benefits of quantized models?<|im_end|>
  <|im_start|>assistant
  """

  Nx.Serving.run(serving, prompt)
```
  Supported FP8 Models

  - Qwen/Qwen3-4B-Instruct-2507-FP8
  - Other FP8 variants following the same weight format

  Notes

[...]

PR #455, opened 2026-05-05
feat: add init cache type opts

polvalente:
This PR adds init_cache types and num_heads for better flexibility of text generation models


lib/bumblebee/text/qwen3.ex
```diff
@@ -373,6 +375,8 @@ defmodule Bumblebee.Text.Qwen3 do
       ],
       query_norm: query_norm,
       key_norm: key_norm,
+      output_hidden_states: false,
+      output_attentions: false,
```
jonatanklosko, line 379: Do we need these new options? We prune these by default, and in order to actually return it in the model output, the user needs to opt-in by configuring global layer options.

lib/bumblebee/layers/transformer.ex
```diff
@@ -416,7 +441,11 @@ defmodule Bumblebee.Layers.Transformer do
       end
 
     {self_attention_cache, cross_attention_cache} =
-      Layers.Decoder.get_attention_caches(block_cache)
+      if is_nil(cross_hidden_state) do
+        {Layers.Decoder.get_self_attention_cache(block_cache), %Axon.None{}}
```
jonatanklosko, line 445: Why do we need a separate function? If cross attention is not enabled then `get_attention_caches` already returns none in the second element (or rather a model that compiles to none).

lib/bumblebee/text/generation.ex
```diff
@@ -313,17 +315,13 @@ defmodule Bumblebee.Text.Generation do
...
-    # the updated cache returned from the model
-    cache =
-      Bumblebee.Utils.Nx.map(cache, fn tensor ->
-        Axon.MixedPrecision.cast(output_policy, tensor, :output)
-      end)
+    # Use the compute precision as the cache type. The key/value tensors are
+    # produced by projection layers running in compute precision, so this
+    # matches what the model will actually return for the cache.
+    cache_type = output_policy.compute || {:f, 32}
+    cache = init_cache(spec, batch_size, max_length, inputs, cache_type: cache_type)
```
jonatanklosko, line 324: Do we get anything from passing it downstream instead of casting as above?

lib/bumblebee/text/generation.ex
```diff
@@ -313,17 +315,13 @@ defmodule Bumblebee.Text.Generation do
...
-    # we make sure the cache we pass as input has the same types as
-    # the updated cache returned from the model
-    cache =
-      Bumblebee.Utils.Nx.map(cache, fn tensor ->
-        Axon.MixedPrecision.cast(output_policy, tensor, :output)
-      end)
+    # Use the compute precision as the cache type. The key/value tensors are
+    # produced by projection layers running in compute precision, so this
+    # matches what the model will actually return for the cache.
+    cache_type = output_policy.compute || {:f, 32}
```
jonatanklosko, line 323: IIRC the cache value returned from attention layers is cast using `:output` precision (since it's the layer output). That's why we cast as output here.

I'm not really sure how to model this with mixed precision policy. It may be that we don't want to cast cache at any point, but then we don't have granularity to specify that, since it's a specific input/output.

lib/bumblebee/text/qwen3.ex
polvalente: I think we can drop these. I must've missed them in my self-review. My focus was on the cache typing

lib/bumblebee/layers/transformer.ex
polvalente: https://github.com/elixir-nx/bumblebee/blob/ccd6f4a7c99d340c3e9f6b5026ecf41f725d10c8/lib/bumblebee/layers/decoder.ex#L218-L224

It always returns cross-attention. Do you mean that cross attention is always Axon.None?

Also, being eager means fewer Axon.nx calls, which reduces the overall graph in Nx.Defn.Evaluator

lib/bumblebee/text/generation.ex
polvalente: I went with using the policy, but originally what I wanted was to introduce a new explicit parameter for the cache type. When I just used the output, at least in my use-case, things ended up using f32 for the cache instead of bf16 like I wanted

lib/bumblebee/text/generation.ex
polvalente: It means we can fit in a smaller memory footprint, if we allocate bf16 instead of f32 and then downcast to bf16

lib/bumblebee/layers/transformer.ex
jonatanklosko: > Do you mean that cross attention is always Axon.None?

Yes!

> Also, being eager means fewer Axon.nx calls, which reduces the overall graph in Nx.Defn.Evaluator

I'd expect that to be marginal, but I am fine either way. If we go with this, I would rather have `put_self_attention_cache`, rather than matching on none in `put_attention_caches`, so it's more aligned on both ends.

lib/bumblebee/text/generation.ex
jonatanklosko: > When I just used the output, at least in my use-case, things ended up using f32 for the cache instead of bf16 like I wanted

Right, my point is that the problem is bigger. We can initialize the cache as bf16, but when predict returns an updated cache it will be cast to f32, as per the output policy, no?

jonatanklosko: On a sidenote, we are using policy from the output node, but technically the attention layers could use a different policy.

polvalente: Can we just add a cache_type option that we can thread through?

polvalente: I'm creating an Axon rewrites library for EMLX and being able to define the cache in a type that won't upcast to f32 is the most important part. I _think_ the rest of this PR can be ignored, but I'm not 100% sure. I'd have to check after removing stuff.

jonatanklosko: Right, but any layer that returns the cache will upcast it to `:f32` if that's the policy `:output` type, unless I am missing something.

polvalente: So you think we should be using the policy output type instead?

jonatanklosko: That's what we do right now, so that cache type going into the model is consistent with the cache type going out of the model. I understand that it is an issue when using compute bf16, and output f32. An actual solution would be to be able to avoid any casting of cache at any point, it's just that I'm not sure how this would look like (this probably means rethinking how the policy and casting works in general, which is also came up in https://github.com/elixir-nx/bumblebee/pull/448#discussion_r2749995535).

jonatanklosko: So I'm fine with passing the output type to cache init. That should work as a memory optimisation, but my understanding is that the issue with compute bf16 and output f32 is still the same :)

polvalente:
Closing for now as I was able to bypass the need for cache size control in my client library

PR #460, opened 2026-06-13
Gemma4 model implementation

edenamram:


lib/bumblebee/text/gemma4_text.ex
```diff
@@ -0,0 +1,966 @@
...
+    initial_state = %{
+      hidden_state: hidden_state,
+      hidden_states: Axon.container({hidden_state}),
+      attentions: Axon.container({}),
+      cache: cache,
+      shared_kv: %{}
+    }
+
+    outputs =
+      Enum.reduce(0..(spec.num_blocks - 1), initial_state, fn idx, state ->
```
jonatanklosko, line 459: We generally use `Layers.Transformer.blocks` as much as possible, to avoid duplicating all that code, which is harder to maintain. When models introduce something novel, we add options to that function as needed to accommodate for the change. Ideally it's something we would do here as well.

jonatanklosko:
Hey, thanks for the PR!

The most important part is to ensure the implementation is correct. [The contributing section](https://github.com/elixir-nx/bumblebee#contributing) describes how to add tests with reference values generated using Python hf/transformers.

Issue #458, opened 2026-06-04
`progress_bar` library no longer actively maintained


petejodo:
As a heads up, the [progress_bar](https://github.com/henrik/progress_bar) library Bumblebee depends on is no longer actively maintained. I discovered this due to updating an unrelated package that depends on `:decimal, "~> 3.0` where `progress_bar` is dependent on v2

AmanTallarium:
Just wanted to +1 this. With many elixir libs updating to a `{:decimal, "~> 3.0"}` requirement (instead of something like `~> 2.0 or ~> 3.0`), for example `ecto_sql`, this is starting to cause problems with resolving deps and therefore preventing package upgrades.

Given there is only a single invocation of the progress_bar in bumblebee, I'd be very happy to implement a simple progress bar inside bumblebee itself (maybe even just a percentage number?) so that the dep can be removed? If that seems like a sensible path forward, please do let me know and I'll work on a PR. Or if there is any other approach I'm happy to help contribute.

josevalim:
Please send a PR with a minimal progress bar and we can evaluate it :)

JustAman62:
Thanks Jose, I've sent over #461 (on my personal account instead of work account), hopefully it's somewhere close enough to a solution.

JustAman62:
The maintainer of progress_bar was looking for new maintainers, so I offered my help and now release v3.1.0 of progress_bar removes the dependency on the decimal lib, so another option to close this issue is to upgrade to decimal 3.1.0.

josevalim:
Thank you for maintaining it. Since bumblebee depends on ~> 3.0, any project that depends on bumblebee can pick up progress_bar 3.1, so this issue is solved. The next question is if we want to remove the progress bar dependency anyway, and that can be evaluated as part of #461. Thank you!

PR #461, opened 2026-06-26
Replace progress_bar with custom utility function

JustAman62:
Removes the `progress_bar` dependency and replaces its single usage with a small utility function which renders a somewhat similar progress bar when downloading models.

Example output snippet (terminal was sized to 50cols to make this easy to read):

```
iex(2)> Bumblebee.load_model({:hf, "google-bert/be
rt-base-uncased"})
|                                    |  1% 4/440MB
|                                    |  2% 8/440MB
|=                                  |  3% 13/440MB
|=                                  |  4% 17/440MB
|=                                  |  5% 22/440MB
|==                                 |  6% 26/440MB
|==                                 |  7% 30/440MB
|==                                 |  8% 35/440MB
|===                                |  9% 39/440MB
|===                                | 10% 44/440MB
```

I've tried to implement just enough for this download use case and not much more, except for making sure the method doesn't failover on weird inputs, and supports non-byte unit progress.

This is my first contribution to this lib so please do let me know if I've forgotten to do something, or am not matching the code style in any way :) 

Relates to #458.


lib/bumblebee/utils/progress_bar.ex
```diff
@@ -0,0 +1,68 @@
...
+  @start_end_char "|"
+  @filled_char "="
+  @unfilled_char " "
+
+  @doc """
+  Renders a simple progress bar to the terminal.
+  The progress bar sizes to fill the entire width of the terminal.
+  """
+  @spec render(number(), number()) :: :ok
+  @spec render(number(), number(), :none | :bytes) :: :ok
```
jonatanklosko: Given that it's optional, let's put `:unit` in options keyword list, similarly to before.

```suggestion
  @spec render(number(), number(), keyword()) :: :ok
```

lib/bumblebee/utils/progress_bar.ex
```diff
@@ -0,0 +1,68 @@
...
+  def render(count, total, unit \\ :none) do
+    percent = min(max(count / total, 0), 1)
+
+    formatted_percent = String.pad_leading("#{trunc(percent * 100)}%", 4)
+    formatted_progress = " " <> formatted_progress(count, total, unit)
+
+    reserved_width =
+      @reserved_width + String.length(formatted_progress) + String.length(formatted_percent)
```
jonatanklosko: We know that the characters used are single byte in utf-8, so we can use `byte_size/1` instead of `String.length/1` to get the length in constant time:

```suggestion
      @reserved_width + byte_size(formatted_progress) + byte_size(formatted_percent)
```

jonatanklosko (commented):
Thanks for the PR, dropping the dependency sounds good to me!

One note: ideally we want the progress bar to update in place in the terminal, similarly to the current behaviour. Likely a matter of prefixing the line with `\r`.

lib/bumblebee/utils/progress_bar.ex
JustAman62: Converted to keyword opts, with a default of `[unit: :none]` to match previous behaviour

lib/bumblebee/utils/progress_bar.ex
JustAman62: Ah very nice, converted to `byte_size`

JustAman62:
Added a carriage return so the progress bar now overwrites the same line. To do this I had to reserve an extra character at the end of the line, because on terminals I was testing on the cursor automatically moved to the next line when the progress bar was written, causing the carriage return to just return to where it is (on the next line). With the extra whitespace, this works consistently now.

```
iex(1)> Bumblebee.load_model({:hf, "google-bert/be
rt-base-uncased"})
|========                         | 25% 110/440MB
```

jonatanklosko:
Thank you :)

